PR #92, opened 2016-10-10
Document `yarn licenses ls`

### LINKED ISSUE #59: yarn licenses ls
Depends on https://github.com/yarnpkg/yarn/issues/530

JoelMarcey:
Fixes #59

### DIFF
--- a/en/docs/cli/licenses.md
+++ b/en/docs/cli/licenses.md
@@ -8,7 +8,44 @@ layout: guide
 
 ##### `yarn licenses ls` <a class="toc" id="toc-yarn-licenses-ls" href="#toc-yarn-licenses-ls"></a>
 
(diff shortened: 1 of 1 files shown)


ahmedelgabri:
> Not sure if this is a bug -- looks like it -- but for packages that have many dependencies and you run on yarn license ls, it non-deterministically stops at random points in the output. e.g., it will list packages starting with a-j, or a-h, or all of them sometimes.

@JoelMarcey are you running the command inside a tmux session? because I have the same issue only inside tmux. But if I pipe it to `pbcopy` for example the output is fine.

Sorry to comment on an old thread.

JoelMarcey:
@ahmedelgabri If I recall correctly, I was running this locally on my Mac, not in any shell session.

PR #105, opened 2016-10-11
Add Aloglia Search

JoelMarcey:
I am not a CSS expert, so if there are any color or positioning changes, we can make modifications as we go along.


JoelMarcey:
screenshot 2016-10-11 08 59 09

PR #106, opened 2016-10-11
Fix post year

adamchainz:
When I first read it I thought I was a bit late to the party!


cpojer:
I was about to do this but then the actual release happened. Thank you.

PR #108, opened 2016-10-11
Fix cache clean docs

pbomb:
`yarn cache clear` does not work, but `yarn cache clean` does. `clear` gives the following error:
```
% yarn cache clear
yarn cache v0.15.1
error Invalid subcommand. Try "ls, clean"
info Visit http://yarnpkg.com/en/docs/cli/cache for documentation about this command.
```


cpojer:
thank you!

PR #121, opened 2016-10-11
Use full key fingerprint in apt-key command

praseodym:
Prevents GPG short key id collision attacks. For more information,
see https://evil32.com.

### DIFF
--- a/en/docs/_installations/linux.md
+++ b/en/docs/_installations/linux.md
@@ -4,7 +4,7 @@ On Debian or Ubuntu Linux, you can install Yarn via our Debian package
 repository. You will first need to configure the repository:
 
 ```sh
-sudo apt-key adv --keyserver pgp.mit.edu --recv 9D41F3C3
+sudo apt-key adv --keyserver pgp.mit.edu --recv 72ECF46A56B4AD39C907BBB71646B01B86E50310
 echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
 ```
 


thejameskyle:
cc @Daniel15

Daniel15:
Out of curiosity, how do you get the full fingerprint? Everything just seems to show the short one.

Would it be safer to `wget` the public key? We already do that for RPM (https://dl.yarnpkg.com/rpm/pubkey.gpg), but that's a totally separate GPG key since RPM doesn't support subkeys.

praseodym:
I used `apt-key adv --list-public-keys --with-fingerprint --with-colons` after adding the key.

NodeSource also uses `wget` to add their key, see e.g. https://deb.nodesource.com/setup_6.x. I think that's a good solution as well.

Daniel15:
[SuperUser](http://superuser.com/a/652073/524623) says to use this command:
```
gpg --keyid-format LONG -k 0x9D41F3C3
```

Which gave me some shorter IDs:
```
pub   4096R/1646B01B86E50310 2016-10-05
uid                          Yarn Packaging <yarn@dan.cx>
sub   4096R/02820C39D50AF136 2016-10-05
sub   4096R/D101F7899D41F3C3 2016-10-05 [expires: 2017-10-05]
```

`D101F7899D41F3C3` seems to work, so I'll update it to use that one. `72ECF46A56B4AD39C907BBB71646B01B86E50310` seems unnecessarily long.

Daniel15:
Updated the site to use the 64-bit key ID (`D101F7899D41F3C3`). Thank you for sending this pull request! I learnt more about GPG 😄

praseodym:
Cheers! With older GPG versions one had to specify the actual fingerprint, not the long key, to be secure against MITM attacks. This should be fine for more recent GPG versions though!

PR #122, opened 2016-10-11
Fix doc link for yarn ls

steveklabnik:
So, either the link is wrong, or the name is wrong. I am pretty sure that this was meant to be for `yarn ls` and not `yarn init`, but let me know if that's incorrect!


steveklabnik:
:confetti_ball:

PR #129, opened 2016-10-11
Update windows.md

wski:


Daniel15:
Also cc @thejameskyle in case he wants this to go somewhere else (but I think the download page is a fine place for it)

PR #136, opened 2016-10-12
docs/cli: Added test command docs.

suitupalex:
The documentation for the `test` command was missing so I decided to add it in with the first commit.

~~In the second commit, I added more information about system-wide commands not being executable by Yarn. Since it's possibly a breaking change for people coming from npm, I felt it should be proactively addressed in the docs.~~

~~The third commit just adds a missing `p.lead` tag to `run`.~~

I tried to lift the wording and formatting from other doc entries.

### DIFF
--- a/_data/guides.yml
+++ b/_data/guides.yml
@@ -108,6 +108,9 @@
   - id: docs_cli_team
     path: /docs/cli/team
     tags: ["cli-team"]
+  - id: docs_cli_test
+    path: /docs/cli/test
+    tags: ["cli-test"]
   - id: docs_cli_unlink
     path: /docs/cli/unlink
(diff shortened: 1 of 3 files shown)


cpojer:
Would you mind reverting the changes to `yarn run`? That seems to me like a bug we should fix in yarn. Let's keep this PR focused on `yarn test`.

addyosmani:
> Let's keep this PR focused on yarn test.

👍

suitupalex:
Removed both the changes to `run` docs and mention of the inability to run system commands. Looks like that issue was found and fixed in yarnpkg/yarn#733.

cpojer:
thank you!

PR #142, opened 2016-10-12
Document that yarn global bin needs to be added to $PATH

Fahrradflucht:
See https://github.com/yarnpkg/yarn/issues/648

### DIFF
--- a/en/docs/cli/global.md
+++ b/en/docs/cli/global.md
@@ -8,7 +8,9 @@ layout: guide
 
 ##### `yarn global` <a class="toc" id="toc-yarn-global" href="#toc-yarn-global"></a>
 
-`yarn global` is a prefix used for a number of commands like `add`, `bin`, `ls` and `remove`. They behave identically to their normal versions except that they use a global directory to store packages. The `global` command makes binaries available to use on your operating system.
+`yarn global` is a prefix used for a number of commands like `add`, `bin`, `ls` and `remove`. They behave identically to their normal versions except that they use a global directory to store packages.
+
+To make the binaries globally available on your operating system you have to add the path returned by `yarn global bin` to your operation systems `$PATH` variable. Note that the path will change on every node version upgrade.
 
 This is useful for developer tooling that is not part of any individual project but instead is used for local commands. One such example is [create-react-app](https://github.com/facebookincubator/create-react-app) which can be installed globally like this:
 


Fahrradflucht:
I thought about adding a recommendation on how to do this ( like adding `export="${PATH}:$(yarn global bin)"` to your shell config, but I think since this works completely different on Windows (IIRC) I omitted that.

cpojer:
cc @kittens – can we make it so we put our global bins into `/usr/local/bin` just like npm does? I think that'll be much easier for people.

Fahrradflucht:
@cpojer wouldn't that have the issues discussed [here](https://github.com/creationix/nvm/issues/668)?

Daniel15:
This is already documented in the installation instructions: https://yarnpkg.com/en/docs/install. See "path setup" on the Mac OS and Linux tabs, and on Windows the installer modifies the path automatically.

Fahrradflucht:
Ah I see. Based on the issue I was referring to, this gets missed sometimes anyway. Do you think I should change it to a little note to make sure that this done on Mac/Linux or do you think it should just be closed all together?

tomchentw:
Thank @Fahrradflucht. I need to add this to my `.zshrc` to make it work on my computer.

To clarify, here's my `npm/node/yarn` install steps:

1. Install `npm` independently with curl script
2. Install `node` with `homebrew` by this command: `brew install node --without-npm`
3. Install `yarn` with `homebrew`: `brew install yarn`

After these steps, my result for `yarn global bin` is:

```
/usr/local/Cellar/node/6.9.1/bin
```

It's not under `~/.yarn/bin` as [the document suggests](https://yarnpkg.com/en/docs/install).

As such, I'll need to modify my `.zshrc` as follows:

```
export PATH="$(yarn global bin):$PATH"
```

Hope someone will find this helpful.

carhartl:
Running ``export PATH="$PATH:`yarn global bin`"`` as suggested by the installation instructions slows down my shell startup noticeably , e.g. time to type. Is there a way to setup the path in a "static" way, while avoiding having to update my dotfiles each time a new node version is installed (`"$PATH:/usr/local/Cellar/node/7.3.0/bin"` => `"$PATH:/usr/local/Cellar/node/7.4.0/bin"`)?

carhartl:
... figured out where to look for: `export PATH="$PATH:$HOME/.config/yarn/global/node_modules/.bin"`

Haroenv:
Sorry for the massive delay on this PR. With the current content this isn't necessary anymore I think. Feel free to reopen and I promise to look at the issues earlier

ianchanning:
I've just hit this issue, there is nothing documenting that you need to add something to your `$PATH` to get `yarn global add` to work. I think this PR is still valid.

The current documentation seems to work around it by specifying that you must use a `prefix` argument. But then libraries such as [Bucklescript installation][1] (and ReasonML) use the default syntax:

```sh
yarn global add bs-platform
```

I raised a [ReasonML issue #2168][2] about the undocumented requirement for adding something to your path for `yarn global add`, but they then claim that it is an issue with yarn.

[1]: https://bucklescript.github.io/docs/en/installation
[2]: https://github.com/facebook/reason/issues/2168

PR #145, opened 2016-10-12
Image Placement bug & optimization

una:
Just some small tweaks. I noticed the image overlapped on the medium media query, so I removed the top margin:

Before:
screen shot 2016-10-12 at 9 54 02 am

After:
screen shot 2016-10-12 at 10 03 50 am

And running the images through an optimizer saved ~ 388kb :)


Daniel15:
Interesting, which optimizer did you use? I already ran the images through TinyPNG in https://github.com/yarnpkg/website/pull/95 (I tried Zopfli but TinyPNG had better results).

una:
@Daniel15 ImageAlpha

thejameskyle:
These images have been run through like 8 image optimizers at this point, I don't understand how they keep getting smaller

Daniel15:
We need to continue optimizing them until there's just a single byte left. That's how it works, right? 😆 

It looks like ImageAlpha has some sort of lossy compression algorithms (https://pngmini.com/lossypng.html), the previous optimizers we've used were probably all lossless.

Issue #133, opened 2016-10-11
front page doesn't mention javascript (or node.js)


ungoldman:
Interesting project!

A bit confusing unless the link to the site is sent in a specific context though.

No mention of what yarn is a package manager *for* on https://yarnpkg.com.

Don't see the words "node" or "javascript" anywhere on https://yarnpkg.com, https://yarnpkg.com/en/docs/getting-started, https://yarnpkg.com/en/docs...

After some searching, the [only](https://github.com/yarnpkg/website/search?l=Markdown&q=javascript&utf8=%E2%9C%93) meaningful mention of javascript on the site is in the title of a blog post: [_posts/2016-10-11-introducing-yarn.md](https://github.com/yarnpkg/website/blob/83b8ee10433e399200340530d60151761caa51b5/_posts/2016-10-11-introducing-yarn.md). Mentions of node in docs here and there, but mainly in reference to needing to install Node.js or talking about files being in `node_modules`.

clentfort:
Buy yarn is not limited to installing JavaScript dependencies. :) At the moment npm is mostly used for JS, but it is not exclusively for JS. Just think about all the CSS and fonts you can install via npm. 

Mentioning node or JavaScript would send a wrong message to the community.

ungoldman:
I see your point @clentfort, but the introductory text on the front page still seems a bit too vague to me. It's a package manager using the npm registry, might be a good idea to at least mention that.

cpojer:
Hey @ungoldman,

thanks for bringing this up. We'll keep the current messaging. Yarn is language agnostic and we'd like to signal that on the website. Nevertheless, you are right, the majority of Yarn usage will be together with JavaScript.

ungoldman:
Maybe mention "language agnostic package manager" on front page?

PR #149, opened 2016-10-12
Adding info that this is a JS package manager

timbotron:
There is no mention that Yarn is a JS package manager on the homepage at all, but at least it should be mentioned in the Getting Started section.


cpojer:
https://github.com/yarnpkg/website/issues/133#issuecomment-253179757

timbotron:
¯\_(ツ)_/¯

PR #155, opened 2016-10-12
Add path setup to installation documentation

jdlehman:
This adds documentation on setting up the path environment variable to provide global access to yarn's bin folder.

### DIFF
--- a/en/docs/_installations/alternatives.md
+++ b/en/docs/_installations/alternatives.md
@@ -20,3 +20,13 @@ Once you have npm installed you can run:
 ```sh
 npm install --global yarn
 ```
+
+### Path Setup
+
+#### Unix/Linux/macOS
+
+{% include_relative _installations/unix_path_setup.md %}
+
+#### Windows
+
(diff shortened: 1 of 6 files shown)


jdlehman:
The Windows section definitely needs some eyes on it. I am not familiar enough with development on a Windows machine.

cpojer:
would you mind rebasing this?

jdlehman:
Not at all!

cpojer:
thank you!

Daniel15:
Both the installer and the Chocolatey package automatically configure the path for you, so you should never have to manually configure it on Windows. I think we should remove this from the Windows tab and just keep it on the "alternatives". Similarly, on Linux, the Yarn executable is installed to `/usr/bin` so it's already on the path.

The only time you need to manually configure the path is if you manually install from a tarball.

Daniel15:
What does this mean by "Yarn's binaries"? Does it mean executables for things you install globally through Yarn itself? 

For Windows, we should make the installer do that automatically, and remove that line from the installation page (as it's confusing) 😄

PR #156, opened 2016-10-12
Flesh out the release process page a bit more

Daniel15:
We'll definitely automate more of this, but I thought I'd at least document the current process.


cpojer:
Thanks Daniel for doing this!

PR #162, opened 2016-10-13
Document `package.json` entry `"flat": true`

wbinnssmith:
This documents the ability to add `"flat": true` as an entry in `package.json`
to enforce the semantics of `yarn install --flat` on every installation, as
implemented in [yarnpkg/yarn#299](https://github.com/yarnpkg/yarn/pull/299). This also notes the drawbacks involved in doing so
(forcing flatness upon consumers of the package).

**This is not ready to merge yet because:**
The only outstanding issue is the link generated using `{{url_base}}` doesn't
include the language, despite it working on other pages. Let me know how to fix
that up and I'll take care of it :)

Also open to rewording things.

### DIFF
--- a/en/docs/package-json.md
+++ b/en/docs/package-json.md
@@ -345,6 +345,18 @@ Optional dependencies can be used with your package, but are not required. If th
 
 Bundled dependencies are an array of package names that will be bundled together when publishing your package.
 
+### `flat` <a class="toc" id="toc-flat" href="#toc-flat"></a>
+
+```json
+{
+  "flat": true
+}
+```
+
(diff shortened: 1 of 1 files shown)


wbinnssmith:
Rendered:

screen shot 2016-10-12 at 8 26 52 pm

thejameskyle:
@wbinnssmith btw when you open a pull request it puts up a demo page in CI so you can just link to it

screen shot 2016-10-13 at 8 22 41 am

cpojer:
ohhh, that's sweet, I didn't realize this. I was checking these out locally.

Great setup, @thejameskyle. The website works so well!

PR #163, opened 2016-10-13
Clarify the role of yarn scripts and note lifecycle scripts

wbinnssmith:
This clarifies the role of scripts, gives (imo) a more concrete example,
and notes lifecycle scripts and when they're invoked.

This started as a simple typo fix and ended up a little more opinionated, and
I'm always happy to reword or change the direction of it.


wbinnssmith:
Rendered:

screen shot 2016-10-12 at 8 55 23 pm

PR #165, opened 2016-10-13
Add NPM version of yarn upgrade

nmn:
This is counter-intuitive, but it works AFAIK.

### DIFF
--- a/en/docs/migrating-from-npm.md
+++ b/en/docs/migrating-from-npm.md
@@ -57,4 +57,4 @@ your existing `npm-shrinkwrap.json` file and check in the newly created `yarn.lo
 | `npm uninstall --save [package]`            | `yarn remove [package]`                     |
 | `npm uninstall --save-dev [package]`        | `yarn remove [package]`                     |
 | `npm uninstall --save-optional [package]`   | `yarn remove [package]`                     |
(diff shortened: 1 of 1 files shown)


cpojer:
Actually I think the table is wrong. `yarn upgrade` doesn't take arguments. `yarn upgrade` updates all packages to the latest version according to their semver range. The equivalent to that in npm is `rm -rf node_modules; npm install`. Would you mind updating the PR?

nmn:
@cpojer I read the documentation, and `yarn upgrade` seems counter-intuitive. It is not used to update dependencies to new versions at all. (I just tried locally)

cpojer:
Yarn upgrade updates dependencies to the latest version based on the semver range, not new versions.

If you have `"jest": "^16.0.0"` in your `package.json` and yarn installs 16.0.0 but 16.1.0 or 16.2.3 comes out, `yarn upgrade` will update to the newer versions. If you have `"jest": "16.0.0"` and a new version of Jest comes out, `yarn upgrade` will stick to 16.0.0.

nmn:
@cpojer Yeah, I didn't say it right. I was trying to use yarn upgrade to change the versions in package.json.

Anyway, fixed in the new commit.

PR #178, opened 2016-10-14
Update windows command to set path

1pete:
shouldn't have a quote in command


Daniel15:
Is the command actually right - Does Yarn really store stuff in `C:\.yarn`? I think it should be the LocalAppData directory...?

1pete:
Right, it depends on your machine, for instance, mine is `C:\Program Files (x86)\Yarn\bin`.

I update it to be more general.

Daniel15:
This won't be needed in the next version (after https://github.com/yarnpkg/yarn/pull/1129 was merged) so I might hold off on this and just totally delete the section instead.

PR #185, opened 2016-10-15
Restore Chocolatey section now that it's approved 🎉🎉🎉

Daniel15:
Chocolatey team approved the package: https://chocolatey.org/packages/yarn/0.15.0


jdalton:
Looks like it's behind the current release *(its 0.15.0 not 0.15.1)*

Daniel15:
Technically, the current release is 0.15.0. 0.15.1 was a patch just for npm and the installer script, all other installation methods are currently still at 0.15.0 as they did not need any changes. The next time we release a new version, everything will be in sync again.

PR #191, opened 2016-10-17
Consistent footer margin

JPeer264:
The footer's margin was not really correct on the `xl` breakpoint. I added a negative margin to the `.footer` class to make the margin consistent.

Furthermore I sorted the breakpoints from `xs` up to `xl` as it is in other files as well.

### DIFF
--- a/_sass/_footer.scss
+++ b/_sass/_footer.scss
@@ -2,6 +2,14 @@
 
 }
 
+@include media-breakpoint-down('xs') {
+  .footer-right {
+    margin-top: 1rem;
+    padding-top: 1rem;
+    border-top: 1px solid $gray-lighter
+  }
+}
+
 @include media-breakpoint-up('sm') {
   .footer-left {
     float: left;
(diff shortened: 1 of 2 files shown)


_sass/_footer.scss
```diff
@@ -29,8 +35,8 @@
   }
 }
 
-@include media-breakpoint-down('md') {
-  .footer-item {
-    display: block;
+@include media-breakpoint-up('xl') {
+  .footer {
+    margin-top: -2.5rem;
```
thejameskyle: What margin are you negating? Can you just update that one?

JPeer264: Actually I just added 

```js
@include media-breakpoint-up('xl') {
  .footer {
    margin-top: -2.5rem;
  }
}
```

The rest was because the breakpoints were not in the right order - that's why git triggered that many changes.

thejameskyle: That's not what I was asking. I'm asking you to modify the `<hr>` that appears above the footer on every page to not have that extra space in the first place so there isn't a need for a negative margin. Just don't modify the hr globally

JPeer264:
I changed now the divider class from the footer to `footer-divider` - I guess you meant that. There is now a consistent margin of `margin: 3rem 0;`

I still ordered the breakpoints.

thejameskyle:
Thanks :)

PR #199, opened 2016-10-20
Update version to 0.16.1 in _config

torifat:
Fixes https://github.com/yarnpkg/yarn/issues/1283

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -6,7 +6,7 @@ baseurl: ""
 url: "https://yarnpkg.com"
 twitter_username: yarnpkg
 github_username:  yarnpkg
-latest_version: 0.16.0
+latest_version: 0.16.1
 gacode: "UA-85522875-1"
 
 exclude:


Daniel15:
cc @bestander - It looks like the 0.16.1 release is incomplete, it lacks the Windows installer and the version number on the site is still the old one.

thejameskyle:
@bestander @Daniel15 What needs to be done with this?

Daniel15:
If we want to bump the versions, we'd need to grab the Windows installer from 0.16.1 and attach it to the release (https://ci.appveyor.com/project/kittens/yarn/build/1.0.244/job/q9y95ox9dtj4cjq3/artifacts) before doing so. The reason Windows installers aren't automatically attached is that we'll Authenticode sign them in the future.

Maybe @bestander knows why 0.16.1 is not fully released.

bestander:
Done, msi uploaded.

bestander:
Sorry guys, I was just not available at the end of the week.
0.16.1 doesn't have any fixes, just a file excluded, so I was not in a
hurry to fix it.
I can try to do it tomorrow

Issue #201, opened 2016-10-20
Add ability to deep link tabs on install/ci instructions


thejameskyle:
Make this work: https://yarnpkg.com/en/docs/install#linux

GreenGeorge:
Would the script to do this be better as a page-specific asset or as a site-wide asset embedded on the layout?

PR #205, opened 2016-10-21
Add tab deep linking

GreenGeorge:
A response to issue #201 
The scripts detects the presence of a hash in the URL and a '.tabs' element on the DOM.
It activates the tabs with the same ID as the hash.
It removes the hash from the URL if there are no matching tabs.
It adds the hash of a tab on the URL bar when a tab is clicked.


thejameskyle:
Thanks, I updated your script to take advantage of the other scripts we're already loading

Issue #179, opened 2016-10-14
Hide language selector on blog 


Devinsuit:
Page https://yarnpkg.com/blog/2016/10/11/introducing-yarn. The link in the language dropdown menu is broken (https://yarnpkg.com/en/2016/10/11/introducing-yarn)

thejameskyle:
The language dropdown should just not show up at all in the blog since it won't be translated

PR #207, opened 2016-10-22
Add a conditional statement to load language selector

GreenGeorge:
This addresses issue #179 
Added a conditional statement to load the language selector in all pages except the blog page.

### DIFF
--- a/_includes/navigation.html
+++ b/_includes/navigation.html
@@ -30,22 +30,24 @@
       </ul>
 
       <ul class="nav navbar-nav pull-lg-right">
-        <li class="nav-item dropdown">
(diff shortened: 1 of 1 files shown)


_includes/navigation.html
```diff
@@ -30,22 +30,24 @@
...
-              {% if language.enabled %}
-                <a href="{{site.baseurl}}/{{language.tag}}/{{vars_url_relative}}"
-                  class="dropdown-item{% if vars_active_language.tag == language.tag %} active{% endif %}">
-                  {{language.name}}
-                </a>
-              {% endif %}
-            {% endfor %}
-          </div>
-        </li>
+        {% unless page.url == '/blog/' %}
```
thejameskyle: This isn't gonna work with subpaths of `/blog/` for posts

GreenGeorge: True. On the next commit I reworked the logic to split the path into an array and check if the 1st element is 'blog'

PR #211, opened 2016-10-24
make background of images transparent

Haroenv:
The images now have a white background, but by replacing it with a transparent background, it is also about 80kb smaller, which will have a small positive impact on the page.

By the way, if you read this Shitty Watercolour, I really like the images


thejameskyle:
The feathering around the colors removes just a little bit too much where you can tell it is cut out. Could you soften and pull it out a bit?

Haroenv:
Sure, I'll redo the changes with a little more feathering. When looking at it side-by-side it's clear that I removed a little too much

Haroenv:
I can't see the difference in the previews now (except that in my version the robot cat doesn't cover the lines, but that's better). The difference in size is now about 70k

thejameskyle:
Thanks!

Haroenv:
With pleasure ❤️

PR #212, opened 2016-10-25
Make clear that publish inherits flags from version

leipert:
See https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/publish.js#L20


leipert:
Seems like noone cared and it got documented here: https://github.com/yarnpkg/website/pull/576

Issue #217, opened 2016-10-25
Remove any docs on bower


thejameskyle:
https://github.com/yarnpkg/yarn/pull/1441

PR #221, opened 2016-10-27
Remove multiple registries featurette

wyze:
This solves #217 as well. This was the only reference to bower, and this bullet was also removed from the readme at [yarnpkg/yarn](https://github.com/yarnpkg/yarn).

@thejameskyle, not sure if you want to address any of the widths of the bottom 2 featurettes now, or leave them as is.


Daniel15:
This was changed to:
> Same Packages
> Install any package from npm and keep your package workflow the same.

PR #222, opened 2016-10-28
Document new --no-git-tag-version option for yarn version

markstos:
The new flag was commited to yarn 10 minutes ago: https://github.com/yarnpkg/yarn/commit/0c63149deddced5c7d4b1fe74fb0d0c496e6f888

Perhaps you wish to wait to merge this until the new version appears in a released version.

### DIFF
--- a/en/docs/cli/version.md
+++ b/en/docs/cli/version.md
@@ -52,3 +52,5 @@ info Current version: 1.0.2
 info New version: 1.0.3
 ✨  Done in 0.09s.
 ```
+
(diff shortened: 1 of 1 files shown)


wyze:
Each flag should have it's own section so it can be linked to. I would modify this to make it look like the `--new-version` flag section.

markstos:
I created PR#223 which addresses @wyze feedback as well as documenting several `config` properties which affect the git behavior of `yarn version --new-version`.  Closing the PR for this version.

PR #227, opened 2016-10-31
fix broken link

goenning:
`url_base` was wrong, thus causing `docs/cli/install#toc-yarn-install-flat` to 404.


thejameskyle:
Thanks

PR #236, opened 2016-11-02
Document new global upgrade comand introduced by ebe899382fdb70222651380f299737c6625d87e4

markstos:


Haroenv:
Thanks!

markstos:
Why didn't you use my commit which also updated `en/docs/cli/global.md`?

Ref: https://github.com/yarnpkg/website/pull/236/commits/a8683bea9d734d04107f4af6f7cc56b1f041c9ea

Haroenv:
That already happened in some other commit, right?

markstos:
Ah, looks like it.

PR #237, opened 2016-11-02
Change http -> https in the installation documentation

erikvold:


thejameskyle:
I think @Daniel15 had a reason for this

Daniel15:
Yeah, two reasons unfortunately:
1. GPG doesn't have a handler for HTTPS URLs installed out-of-the-box:
```
gpgkeys: protocol 'https' not supported
gpg: no handler for keyserver scheme 'https'
gpg: WARNING: unable to fetch URI https://dl.yarnpkg.com/debian/pubkey.gpg: keyserver error
```

2. The GnuTLS version in older releases of Ubuntu (12.04 and 14.04) doesn't like CloudFlare SSL certs (see #607):
```
Err https://dl.yarnpkg.com stable/main amd64 Packages
  gnutls_handshake() failed: Handshake failed
Err https://dl.yarnpkg.com stable/main i386 Packages
  gnutls_handshake() failed: Handshake failed
```

Let's Encrypt certs seem fine, but @thejameskyle mentioned in #607 that he's already getting a cert set up. We can change the repo URL once we get a proper SSL cert, but unfortunately I don't think we can change the GPG one to HTTPS. We can probably change it back to using a keyserver, we just need to find a server that's reliable. Previously we were using `pgp.mit.edu` but it had sporadic outages.

Going to close this, but let's revisit once we get a proper SSL cert.

Thanks for trying Yarn! 😄

thejameskyle:
@Daniel15 There is a private cert setup now

Daniel15:
Ahh... I'll retest Ubuntu 12.04 and see if https works.

shadone:
Hi there. Any update for this issue? GnuPG indeed doesn't support fetching from HTTPS url however the common workaround is to use curl and feed the data to GnuGP or better apt-key like so `curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -`

Daniel15:
Hey @shadone  - The site was already updated to use HTTPS for the GPG key. I just tested with older Ubuntu versions and HTTPS seems to work fine for the repo, so I updated the site too: e6761160fef64f89e31749592bb14490ae8e9df2

PR #242, opened 2016-11-03
Document new short-flags for `add`

markstos:


markstos:
Oops, I just noticed PR#240 also implements this. Now there are choices for 2 different documentation styles. :)

PR #245, opened 2016-11-04
Document nightly builds on download page

Daniel15:
References https://github.com/yarnpkg/yarn/issues/1474

### DIFF
--- a/en/docs/_installations/nightly.md
+++ b/en/docs/_installations/nightly.md
@@ -1,0 +1,19 @@
+### Nightly Builds
+
+Nightly builds are the latest and greatest versions of Yarn, built using the very latest Yarn source code. Nightly builds are useful to try new features or test bug fixes that have not yet been released as part of a stable release. However, these builds are **not guaranteed to be stable** and may have bugs. See [the nightly build site](https://nightly.yarnpkg.com/) for more information.
+
+The easiest way of installing a nightly build is via our shell script:
+```sh
+wget https://yarnpkg.com/install.sh
+chmod +x install.sh
+./install.sh --nightly
+```
+
+A Ubuntu/Debian repository of the nightly builds is also available. To enable it, run the following commands:
+```sh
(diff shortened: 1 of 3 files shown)


en/docs/install.html
```diff
@@ -15,6 +15,7 @@
 {% capture windows      %}{% include_relative _installations/windows.md      %}{% endcapture %}
 {% capture linux        %}{% include_relative _installations/linux.md        %}{% endcapture %}
 {% capture alternatives %}{% include_relative _installations/alternatives.md %}{% endcapture %}
+{% capture nightly      %}{% include_relative _installations/nightly.md      %}{% endcapture %}
```
thejameskyle: Do you think we should emphasize nightly builds so much? I feel like we should make it a separate thing

Daniel15: Like a totally separate page? It seemed to fit in nicely here and I imagine this is where people would look for the info, I can put it elsewhere if you like though!

thejameskyle: It doesn't need to be a separate page, but maybe it's own section on the same page. I don't know how well "nightly" is going to translate and I don't want people installing it because they weren't able to use one of the other options

thejameskyle: I'm imagining a separate box that explains what nightly is that can be expanded to include instructions

en/docs/_installations/nightly.md
```diff
@@ -0,0 +1,17 @@
+The easiest way of installing a nightly build is via our shell script:
+```sh
+wget https://yarnpkg.com/install.sh
+chmod +x install.sh
+./install.sh --nightly
+```
+
+A Ubuntu/Debian repository of the nightly builds is also available. To enable it, run the following commands:
```
Daniel15: Thanks, I can change that!

Daniel15:
@thejameskyle  - Are there any other changes you'd like, or is this good to merge now (apart from changing "a Ubuntu" to "an Ubuntu")?

Daniel15:
Going to merge this and iterate (and/or ask for forgiveness) if we don't like this approach.

PR #248, opened 2016-11-05
Remove reference to npm

semitrivial:
The yarn website contains a mention of `npm init` which should probably be changed to `yarn init`, this PR fixes that.


thejameskyle:
lol

Issue #181, opened 2016-10-14
Clarify what' yarn ls' is showing


swang:
_Reposting from yarnkpkg/yarn_

`yarn ls` has no clear documentation. I would like to add some but it I want to be clear what is going on.

![modules](https://cloud.githubusercontent.com/assets/49867/19367199/01aeeae2-914f-11e6-91a9-54085c10929a.png)
From the screenshot, it looks like zip-stream requires a versioning of lodash@~3.2.0 in grey, while the bolded **lodash@3.2.0** is the actual version of lodash that was installed.

But.. looking into the code, the grey/"dim" npm modules that don't have a bold equivalent (ex: compress-commons) means that they have been hoisted up to the main `node_modules/` folder. So anything that is bolded, is actually dependencies that yarn decided it couldn't hoist out to the main folder. Is this mostly correct? Is there something else that should be known in the documentation?

olingern:
My comment in [https://github.com/yarnpkg/yarn/issues/1123](https://github.com/yarnpkg/yarn/issues/1123) may help.

markstos:
It's also not clear what `yarn ls --depth 0` is showing. I would expect it to list the same dependencies seen in `package.json`, but it lists many more.  I suspect it's listing nested dependencies which got flattened to the top level, even though they are not top-level dependencies. This behavior should be documented.

markstos:
This should be re-opened, it's still not clear what `yarn list --depth 0` is showing. See: https://github.com/yarnpkg/website/pull/251#pullrequestreview-75150705

olingern:
@markstos I'll provide a detail explanation of what is being displayed here later today.

markstos:
@olingern It does seem like `yarn list --depth 0` is showing flattened dependencies. I guess if you want the un-flattened dependencies, you can just open up `package.json`. If the current behavior is considered desired, then appropriate mentions of flattening should be added to the `yarn list` docs.

hgezim:
@markstos `package.json` doesn't actually show you what's installed, though.

What does bold mean in the `yarn ls` output?

PR #251, opened 2016-11-08
Add documentation for ls and ls --depth command

olingern:
Documentation for added functionality at [#1463](https://github.com/yarnpkg/yarn/pull/1463)

### DIFF
--- a/en/docs/cli/ls.md
+++ b/en/docs/cli/ls.md
@@ -12,6 +12,10 @@ layout: guide
 yarn ls
 ```
 
+The `yarn ls` command mimics the expected Unix behavior of listing. In Yarn, the `ls` 
+command lists all dependencies for the current working directory by referencing all 
+package manager meta data files, which includes a project's dependencies.
+
 ```
 yarn ls vx.x.x
 ├─ package-1@1.3.3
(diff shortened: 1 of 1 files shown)


dhruvdutt:
Should close #181

dhruvdutt:
Please have a look at #181

olingern:
@dhruvdutt My comment in [https://github.com/yarnpkg/yarn/issues/1123](https://github.com/yarnpkg/yarn/issues/1123#issuecomment-258986245) may help.

dhruvdutt:
Thanks for the reference and your efforts documenting it.

en/docs/cli/ls.md
```diff
@@ -4,18 +4,32 @@ guide: docs_cli
 layout: guide
 ---
 
-<p class="lead">List installed packages.</p>
+<p class="lead">List all dependencies installed for the current working directory</p>
```
thejameskyle: I don't think you need to specify current working directory in the description of this command.

olingern: I added it in case yarn decided to add support for a `--global` flag in the future, differentiating the default behavior from a global listing. I could also see someone very new not understanding `ls` is relative to the cwd and the directory's package.json / bower.json / yarn.json. I think without the 'cwd' clause, it assumes the user has some knowledge about `ls` and basic unix commands.

thejameskyle: That does exist as [`yarn global ls`](https://yarnpkg.com/en/docs/cli/global), I don't see Yarn adding `yarn ls --global`

olingern: I see. Makes sense. Updated and rebased!

PR #256, opened 2016-11-11
Use https for debian public key link

eins78:


Daniel15:
GPG doesn't have a handler for HTTPS URLs installed out-of-the-box:
```
gpgkeys: protocol 'https' not supported
gpg: no handler for keyserver scheme 'https'
gpg: WARNING: unable to fetch URI https://dl.yarnpkg.com/debian/pubkey.gpg: keyserver error
```

However, we should be able to use cURL. Could you please try something like this:

```
curl https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
```

Daniel15:
Fixed by https://github.com/yarnpkg/website/commit/ba516f0478f7dcb279bc31ffdaea2dc8fa6a2885

PR #260, opened 2016-11-13
Move nightly builds page to site

### LINKED ISSUE #259: Merge in nightly builds page
https://nightly.yarnpkg.com/

@Daniel15 Any reason this can't be a page on the normal website? The styles and layout have already gone out of sync.

If there's something dynamic you could just turn it into a JSON api that we can fetch from.

Daniel15:
Moves the page at https://nightly.yarnpkg.com/ onto the main site. The list of available builds is loaded via an AJAX request from the nightly build server.

Once this lands, I'll make https://nightly.yarnpkg.com/ redirect to it and kill that old page. It was mainly for prototyping some ideas 😄 

cc @thejameskyle 

Closes #259

### DIFF
--- a/_data/guides.yml
+++ b/_data/guides.yml
@@ -9,8 +9,6 @@
     path: /docs/install
   - id: docs_usage
     path: /docs/usage
-  - id: docs_nightly
-    path: /docs/nightly
 
 - id: docs_yarn_workflow
   title: docs_yarn_workflow_title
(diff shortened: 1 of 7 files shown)


thejameskyle:
I'm not sure this should be in the getting started guide. I'm tempted to just leave it as a link on the install page unless there's a better place for it

Daniel15:
I wasn't sure where to put it. Should I just remove the link from ` _data/guides.yml`?

thejameskyle:
Yeah I think that would be best, eventually I'll do a slight reorganization of some stuff and include it in another guide

thejameskyle:
Awesome work on this page btw, it looks really great

PR #268, opened 2016-11-16
Add information about installing specific version of Yarn on Travis and Circle

Daniel15:
Now that our Debian repo can contain multiple version of Yarn (implemented in https://github.com/yarnpkg/yarn/issues/1839), we should advocate locking the version of Yarn used in CI environments. The main reason is so that people's builds don't suddenly break because they're always installing the latest version of Yarn (which could have breaking changes for their environment) 😛

### DIFF
--- a/en/docs/_ci/deb-specific-version.md
+++ b/en/docs/_ci/deb-specific-version.md
@@ -1,0 +1,4 @@
+It is recommended that you lock in a specific version of Yarn, so that all your builds use the same version of Yarn, and you can test new Yarn releases before switching over. You can do this by adding the version number to the `apt-get install` call:
+```
+sudo apt-get install -y -qq yarn={{site.latest_version}}-1
+```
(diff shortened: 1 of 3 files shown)


en/docs/_ci/deb-specific-version.md
```diff
@@ -0,0 +1,4 @@
+It is recommended that you lock in a specific version of Yarn, so that all your builds use the same version of Yarn, and you can test new Yarn releases before switching over. You can do this by adding the version number to the `apt-get install` call:
+```
+sudo apt-get install -y -qq yarn={{site.latest_version}}-1
```
Daniel15: All our Debian packages have `-1` at the end of the version (eg. `0.16.0-1`). The idea is that it's bumped (eg. to `0.16.0-2`) if you just make changes to the .deb file without modifying the software, like if you need to rebuild it for some reason.

PR #270, opened 2016-11-17
Remove redundant sudo from Debian install instructions

drakedevel:
There's no need to run curl as root if it's writing the contents to standard output. Verified that the new command works just as well on my Debian system.

### DIFF
--- a/en/docs/_installations/linux.md
+++ b/en/docs/_installations/linux.md
@@ -4,7 +4,7 @@ On Debian or Ubuntu Linux, you can install Yarn via our Debian package
 repository. You will first need to configure the repository:
 
 ```sh
-sudo curl https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
 echo "deb http://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
 ```
 


thejameskyle:
cc @Daniel15

Daniel15:
So, I had a strange issue when testing. The problem I had was that if `sudo` is prompting for a password, piping the result of `curl` into `apt-key` *appears* to hang. In reality, it's sitting waiting for you to enter your password. Sometimes something eats the prompt (perhaps cURL is rendering over it?), and other times it appears in a strange place:


Using `sudo` for the `curl` prompts for your password upfront *before* doing anything, which means the prompt will never appear when running `apt-key` and everything works fine:


So it's a bit of a hack, but not *too* bad.

Another option is to do this:
```
sudo sh -c 'curl https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -'
```

I don't think that's significantly better though.

Daniel15:
Silent mode (`curl -s`) might be fine too, it'd also avoid this issue.

drakedevel:
Hm, that's an unfortunate interaction.

I think my preferred solution is `curl -sS` -- `-s` alone silences error messages which would otherwise be valuable. I confirmed this option is present in Debian at least since Squeeze.

`wget` is actually more verbose by default (at least in Jessie where I tested):
```
adrake@4022e5a68695:/$ wget -O- https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add
[sudo] password for adrake: converted 'https://dl.yarnpkg.com/debian/pubkey.gpg' (ANSI_X3.4-1968) -> 'https://dl.yarnpkg.com/debian/pubkey.gpg' (UTF-8)
--2016-11-17 20:57:19--  https://dl.yarnpkg.com/debian/pubkey.gpg
Resolving dl.yarnpkg.com (dl.yarnpkg.com)... 104.16.63.173, 104.16.59.173, 104.16.60.173, ...
Connecting to dl.yarnpkg.com (dl.yarnpkg.com)|104.16.63.173|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 7453 (7.3K) [application/octet-stream]
Saving to: 'STDOUT'

-                                   100%[==================================================================>]   7.28K  --.-KB/s   in 0.002s 

2016-11-17 20:57:19 (4.26 MB/s) - written to stdout [7453/7453]
```

One minor point in favor of `wget` is that it is installed by default with the Debian minimal system as it is `Priority: important`, whereas `curl` is `Priority: optional`. Unfortunately there does not appear to be any way to suppress non-error output, `-q` suppresses error messages and `-nv` prints a non-error status line to stderr.

Daniel15:
> One minor point in favor of wget is that it is installed by default with the Debian minimal system as it is Priority: important, whereas curl is Priority: optional

Good point. cURL also has more dependencies than wget. Perhaps we should switch it to use wget in that case? I imagine wget has a silent flag too.

drakedevel:
It does have a silent flag (`-q`) but as I mentioned at the end of my last comment, it suppresses all error messages (admittedly, the target audience can probably figure out to drop the flag to debug). Happy to update this PR if you want to go that route.

Daniel15:
Ahh, my mistake, I totally missed that part of your comment! 😄

PR #272, opened 2016-11-17
Clarify install documentation

elwayman02:
It's unexpected that the `yarn install` command would not respect the versions listed in `package.json`. That's less than ideal, in my personal opinion, as I'd prefer to just run `yarn` every time I update my `package.json` to get the new versions installed. However, if this is in fact the intended behavior, we should at least update the documentation to make it clear that this is the behavior, since the current documentation explicitly describes the behavior I expected, which is not what was implemented.

### DIFF
--- a/en/docs/cli/install.md
+++ b/en/docs/cli/install.md
@@ -20,8 +20,8 @@ Running `yarn` with no command will run `yarn install`, passing through any prov
 
 ##### `yarn install` <a class="toc" id="toc-yarn-install" href="#toc-yarn-install"></a>
 
-Install all the dependencies listed within `package.json` in the local
-`node_modules` folder.
+Install all the dependencies listed within `yarn.lock` in the local
+`node_modules` folder. Note: `package.json` will be used if `yarn.lock` is not present.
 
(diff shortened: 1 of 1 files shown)


thejameskyle:
That's not the behavior of Yarn, if you update the `package.json` it invalidates the `yarn.lock`.

elwayman02:
@thejameskyle I verified before creating this PR that Yarn v0.16.1 does *not* invalidate the `yarn.lock` when you run `yarn install`, only when you run `yarn update`. Can you please clarify why you think it does? I'm really confused now, especially since @rwjblue told me the exact opposite of what you just said.

thejameskyle:
I just verified this behavior myself on 16.1 and it does invalidate.

```sh
yarn add lodash@2
yarn outdated
```

```
yarn outdated v0.16.1
Package Current Wanted Latest
lodash  2.4.2   2.4.2  4.17.2
```

Edit `package.json` for lodash@4:

```sh
yarn install
yarn outdated
```

```
yarn outdated v0.16.1
✨  Done in 0.50s.
```

elwayman02:
What happens if you manually change the `lodash` version in package.json instead of using `yarn add`?

PR #273, opened 2016-11-18
Add docs for using the `--version` flag with the installation script

naganowl:
Document how to use the changes per https://github.com/yarnpkg/yarn/pull/1928


naganowl:
Sorry about that! It was more of a placeholder and a bit rushed, thanks for catching that. I've modified it to keep the inline usage of it as I figured the manual example was right underneath making it apparent how it could be used as in your example.

I originally linked to the releases, but thought that the tags were more concise to see the versions. I've made the change to releases following your explanation though.

PR #274, opened 2016-11-18
Refer to `yarn global bin` in path setup

MarkBennett:
When installing with Homebrew or other tools, yarn doesn't always use `~/.yarn/bin`. Instead direct users to use `yarn global bin` to refer to the correct path.


thejameskyle:
@Daniel15

Daniel15:
Looks good to me! Thank you 😄

PR #275, opened 2016-11-18
Latest version is 0.17.4

swrobel:
Version on the site is quite out of date. See releases page: https://github.com/yarnpkg/yarn/releases

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -6,7 +6,7 @@ baseurl: ""
 url: "https://yarnpkg.com"
 twitter_username: yarnpkg
 github_username:  yarnpkg
-latest_version: 0.16.1
+latest_version: 0.17.4
 gacode: "UA-85522875-1"
 
 exclude:


IanVS:
Just a suggestion, but it would probably be nice to automate this, to avoid needing to update it every time there's a new release.  I was also a bit confused by seeing an out-of-date version.  IMHO not displaying a version at all would be better than an out-of-date version.

Daniel15:
This is intentional. 0.17.x is on hold at the moment, until some launch blocking bugs are fixed.

> Just a suggestion, but it would probably be nice to automate this

@IanVS  - Yes, this will be automated, but right now we're intentionally holding back on 0.16.1 due to bugs in 0.17.x. 0.17.x will officially launch soon - Probably today or tomorrow. You can follow along in https://github.com/yarnpkg/yarn/issues/1859.

swrobel:
@Daniel15 the 0.17.x versions are showing up in Homebrew...

IanVS:
And the version returned by `npm info yarn` is `0.17.6`.

Daniel15:
We released Yarn 0.17.6 as stable earlier today :)

The source of truth for the latest version of Yarn is
https://yarnpkg.com/latest-version, which pulls from this config file.

PR #280, opened 2016-11-20
Add documentation for --mutex

### LINKED ISSUE #261: Document `--mutex` option
We're using yarn on our CI because it is awesome. But we're getting strange cryptic errors about files in `~/.yarn-cache/` throwing JSON.parse errors and it's rarely the same error twice.

The solution was to use `yarn --mutex file` (or `yarn --mutex network`), but it's [barely documented](https://github.com/yarnpkg/yarn/issues/1303) anywhere ([outside of the source code](https://github.com/yarnpkg/yarn/blob/c209f84b8ab160fa711916b753f2b983daa21b9c/src/cli/index.js#L366)), and [not at all on the website](https://www.google.com.au/search?q=mutex+site%3Ayarnpkg.com).

grrowl:
Adds a note in `cli/index.md` describing usage of the `--mutex` flag in both file and network modes.

Fixes #261

### DIFF
--- a/en/docs/cli/index.md
+++ b/en/docs/cli/index.md
@@ -20,7 +20,7 @@ While all of the available commands are provided here, in alphabetical order, so
 
 Running `yarn` with no command will run `yarn install`, passing through any provided flags.
 
-## Concurrency and `--mutex` <a class="toc" id="toc-default-command" href="#toc-concurrency-and---mutex"></a>
+## Concurrency and `--mutex` <a class="toc" id="toc-concurrency-and-mutex" href="#toc-concurrency-and-mutex"></a>
 When running multiple instances of yarn as the same user on the same server,
 you can ensure only one instance runs at any given time (and avoid conflicts)
 by passing the global flag `--mutex` followed by `file` or `network`. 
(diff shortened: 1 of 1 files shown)


en/docs/cli/index.md
```diff
@@ -19,3 +19,7 @@ While all of the available commands are provided here, in alphabetical order, so
 ## Default Command <a class="toc" id="toc-default-command" href="#toc-default-command"></a>
 
 Running `yarn` with no command will run `yarn install`, passing through any provided flags.
+
+## Concurrency and --mutex
+
+When running multiple instances of yarn as the same user on the same server, you can ensure only one instance runs at any given time (and avoid conflicts) by passing the global flag `--mutex (file|network)`. This will write/read a mutex file `.yarn-single-instance` in the current working directory or open network port 31997. You can specify an alternate or global filename, or an alternate network port by specifying `--mutex file:/tmp/.yarn-mutex` or `--mutex network:31999`.
```
thejameskyle: It'd be better to break these apart into separate examples of how to use each part of this flag.

grrowl:
💯 so much better. Almost considered moving the "by default" paragraphs into `# comments` within the sh code examples, but if you don't mind the extra paragraphs it's definitely clearer as non-monospaced english.

thejameskyle:
I prefer not putting text into code because that makes it harder to translate

PR #285, opened 2016-11-23
Removing inaccurate line from the migration docs

silverlight513:
`yarn upgrade` is not the same as `rm -rf node_modules/ && npm install`. The npm install command doesn't upgrade dependancies to their latest

### DIFF
--- a/en/docs/migrating-from-npm.md
+++ b/en/docs/migrating-from-npm.md
@@ -59,4 +59,3 @@ your existing `npm-shrinkwrap.json` file and check in the newly created `yarn.lo
 | `npm uninstall --save-dev [package]`        | `yarn remove [package]`                     |
 | `npm uninstall --save-optional [package]`   | `yarn remove [package]`                     |
 | `npm cache clean`                           | `yarn cache clean`                          |
-| `rm -rf node_modules && npm install`        | `yarn upgrade`                              |


Daniel15:
The [documentation for `yarn upgrade`](https://yarnpkg.com/en/docs/cli/upgrade) says:
> This command updates all dependencies to their latest version based on the version range specified in the package.json

Isn't that what `npm install` does when `node_modules` is empty? The docs seem correct to me.

silverlight513:
`npm install` doesn't upgrade anything. It simply installs the packages according to the version numbers in your package.json. The same as yarn install. 

When running `rm -r node_modules && npm install` the npm shrinkwrap doesn't update because no dependancies are updated. When running `yarn upgrade` the yarn lock file is updated because updates were made. 

When you simply want to refresh your node modules directory and nothing else, the `yarn upgrade` command doesn't do this.

Daniel15:
> the npm shrinkwrap doesn't update because no dependancies are updated

This assumes you're using npm shrinkwrap - Most npm users aren't using shrinkwrap due to a number of issues with it.

> npm install doesn't upgrade anything. It simply installs the packages according to the version numbers in your package.json

My understanding is similar to what `yarn update` does. If there's a newer package version that matches the range in `package.json` but is **not** installed yet, it'll update to that version. That's the same behaviour as deleting `node_modules` then running `npm install` - You'll get the latest version that matches the range in package.json.

silverlight513:
If it does indeed stay true to what the package.json specifies then please close this PR. 

My main reason for it was constantly putting up pointless commits of a yarn.lock update because it kept changing when I didn't want it to. I still don't believe it is an exact match because it does update a file but it's close enough.

PR #286, opened 2016-11-23
Fix CircleCI instruction

torifat:
Without `override`, CircleCI will run `npm install` instead of `yarn install`


Daniel15:
Good catch, thanks!

PR #288, opened 2016-11-28
Clarify the symbol before commit-ish

DenisGorbachev:
See https://github.com/yarnpkg/yarn/issues/1988

### DIFF
--- a/en/docs/cli/add.md
+++ b/en/docs/cli/add.md
@@ -48,7 +48,7 @@ You can also specify packages from different locations:
 3. `yarn add file:/path/to/local/tarball.tgz` installs a package from a gzipped
   tarball which could be used to share a package before publishing it.
 4. `yarn add <git remote url>` installs a package from a remote git repository.
-5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote git repository at specific git commit or git tag.
+5. `yarn add <git remote url>#<commit/tag>` installs a package from a remote git repository at specific git commit or git tag.
 6. `yarn add https://my-project.org/package.tgz` installs a package from a
   remote gzipped tarball.
 
--- a/en/docs/cli/add.md
+++ b/en/docs/cli/add.md
@@ -48,7 +48,7 @@ You can also specify packages from different locations:
 3. `yarn add file:/path/to/local/tarball.tgz` installs a package from a gzipped
   tarball which could be used to share a package before publishing it.
 4. `yarn add <git remote url>` installs a package from a remote git repository.
-5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote git repository at specific commit.
+5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote git repository at specific git commit or git tag.
 6. `yarn add https://my-project.org/package.tgz` installs a package from a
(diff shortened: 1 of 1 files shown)


DenisGorbachev:
@Harrisonbro sure, added

Daniel15:
This seems a little redundant since we list Git twice - Once with the commit-ish and once without:
> 4. `yarn add <git remote url>` installs a package from a remote git repository.
> 5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote git repository at specific git commit or git tag.

Perhaps we could instead combine them into one, and say something like:
> 5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote Git repository. `commit-ish` is optional and can be used to refer to a specific git commit or git tag.

Harrisonbro:
That's true, @Daniel15, though be aware that the listing for package versions has just as much redundancy:

> You can specify versions using one of these:
> 
> 1. `yarn add package-name` installs the "latest" version of the package.
> 2. `yarn add package-name@1.2.3` installs a specific version of a package from
  the registry.
> 3. `yarn add package-name@tag` installs a specific
  ["tag"]({{url_base}}/docs/cli/tag) (e.g. `beta`, `next`, or `latest`).

Personally, I'd argue that in both cases the additional example makes it clearer to the reader. However, given that using Git URLs is a rarer case I see the point about making it more concise. 

If we do the more concise version I would suggest we clarify what will be installed if `commit-ish` is omitted. I'd suggest this edited version:

> 5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote Git repository. `#<commit-ish>` is optional and tells Yarn to install a specific commit or Git tag; if omitted, the latest commit on the remote's `master` branch will be installed.

_Note: I don't actually know what Yarn will do if you just give a Git URL, so someone should verify whether what I say above is true._

DenisGorbachev:
I vote for two examples, since @Harrisonbro was confused initially.

Who's going to make the decision?

en/docs/cli/add.md
```diff
@@ -48,7 +48,8 @@ You can also specify packages from different locations:
 3. `yarn add file:/path/to/local/tarball.tgz` installs a package from a gzipped
   tarball which could be used to share a package before publishing it.
 4. `yarn add <git remote url>` installs a package from a remote git repository.
-5. `yarn add https://my-project.org/package.tgz` installs a package from a
+5. `yarn add <git remote url>#<commit-ish>` installs a package from a remote git repository at specific git commit or git tag.
```
thejameskyle: We should just `commit/tag` instead of `commit-ish`

Could we just modify 4 to be `yarn add <git remote url>(#<commit/tag>)` or is that more confusing.

thejameskyle: I should really read threads before commenting...

Daniel15:
> Who's going to make the decision?

I'll leave that up to @thejameskyle, he's the webmaster 😛

DenisGorbachev:
@thejameskyle so should we change anything or not?

thejameskyle:
Just replace `commit-ish` with `commit/tag`

DenisGorbachev:
@thejameskyle done - could you please merge?

PR #290, opened 2016-11-29
Move installation script to site rather than living in the Yarn repo

Daniel15:
This makes more sense to live on the site or in the releases repo, as it's not a part of Yarn itself. This is similar to what other projects with installation scripts do - The installation script lives on the site.

References https://github.com/yarnpkg/yarn/issues/2058

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -29,6 +29,7 @@ exclude:
 
 include:
   - _redirects
+  - install.sh
 
 gems:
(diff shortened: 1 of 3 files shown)


Daniel15:
Confirmed it loads fine in the site preview: 
```
wget https://deploy-preview-290--assembler-pig-23305.netlify.com/install.sh
chmod +x install.sh
./install.sh
```

PR #291, opened 2016-11-29
Add note about why installing via npm is not recommended

Daniel15:
I tried to be as neutral as possible while writing this, let me know if it doesn't make sense.

### DIFF
--- a/en/docs/_installations/alternatives.md
+++ b/en/docs/_installations/alternatives.md
@@ -11,13 +11,13 @@ recommended to install Yarn via our packages instead.
 
 #### Install via npm
 
-> Note: Installation via npm is generally not recommended. npm is 
-> non-determinstic, packages are not signed and npm does not perform any 
+> **Note:** Installation via npm is generally not recommended. npm is 
+> non-determinstic, packages are not signed, and npm does not perform any 
 > integrity checks other than a basic SHA1 hash, which is a security risk 
 > when installing system-wide apps.
 >
-> For these reasons, It is highly recommended that you install Yarn through the
-installation method best suited to your operating system.
+> For these reasons, it is highly recommended that you install Yarn through the
+> installation method best suited to your operating system.
 
 You can also install Yarn through the [npm package manager](http://npmjs.org/)
 if you already have it installed. If you already have
--- a/en/docs/_installations/alternatives.md
+++ b/en/docs/_installations/alternatives.md
@@ -11,12 +11,10 @@ recommended to install Yarn via our packages instead.
 
 #### Install via npm
 
-> Note: Installation via npm is generally not recommended. Installing via npm is
-significantly slower than all of our other installation methods. npm packages
-are not signed and npm does not perform any integrity checks, which is a
-security risk when installing system-wide apps. Additionally, in an environment
-with a system-wide package manager such as most Linux distributions, apps
-installed via npm are not updated when you update the rest of your system.
+> Note: Installation via npm is generally not recommended. npm is 
+> non-determinstic, packages are not signed and npm does not perform any 
+> integrity checks other than a basic SHA1 hash, which is a security risk 
+> when installing system-wide apps.
 >
 > For these reasons, It is highly recommended that you install Yarn through the
 installation method best suited to your operating system.
--- a/en/docs/_installations/alternatives.md
+++ b/en/docs/_installations/alternatives.md
@@ -11,6 +11,16 @@ recommended to install Yarn via our packages instead.
 
 #### Install via npm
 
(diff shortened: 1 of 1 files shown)


en/docs/_installations/alternatives.md
```diff
@@ -11,6 +11,16 @@ recommended to install Yarn via our packages instead.
 
 #### Install via npm
 
+> Note: Installation via npm is generally not recommended. Installing via npm is
+significantly slower than all of our other installation methods. npm packages
+are not signed and npm does not perform any integrity checks, which is a
+security risk when installing system-wide apps. Additionally, in an environment
+with a system-wide package manager such as most Linux distributions, apps
+installed via npm are not updated when you update the rest of your system.
```
thejameskyle: The only really convincing bit here is the security check, I would simplify it to just that.

Daniel15:
I should probably also mention the fact that dependency versions are locked down with all other installation methods, so things like https://github.com/yarnpkg/yarn/issues/2072 are avoided.

en/docs/_installations/alternatives.md
palmerj3: Is this true? Fairly certain NPM DOES do shasum validation - https://github.com/npm/npm/blob/latest/lib/cache/add-remote-tarball.js#L86

Daniel15:
Hash verification is not the same. It doesn't verify **who** uploaded the
package or that it's actually a legitimate release, it's just a hash of the
contents. Additionally, npm uses SHA1, which is not considered secure.

Sent from my phone.

On Dec 2, 2016 4:00 AM, "Jason Palmer" <notifications@github.com> wrote:

*@palmerj3* commented on this pull request.
------------------------------

In en/docs/_installations/alternatives.md
<https://github.com/yarnpkg/website/pull/291>:

> @@ -11,6 +11,16 @@ recommended to install Yarn via our packages instead.

 #### Install via npm

+> Note: Installation via npm is generally not recommended. Installing
via npm is
+significantly slower than all of our other installation methods. npm packages
+are not signed and npm does not perform any integrity checks, which is a
+security risk when installing system-wide apps. Additionally, in an environment
+with a system-wide package manager such as most Linux distributions, apps
+installed via npm are not updated when you update the rest of your system.

Is this true? Fairly certain NPM DOES do shasum validation -
https://github.com/npm/npm/blob/latest/lib/cache/add-remote-tarball.js#L86

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/yarnpkg/website/pull/291>, or mute the thread

wjordan:
> I should probably also mention the fact that dependency versions are locked down with all other installation methods, so things like yarnpkg/yarn#2072 are avoided.

@Daniel15 floating dependency versions are not required in `package.json`, it's just an `npm install` default. You really should lock dependency versions in `package.json`, if its important to keep `yarn`'s correct behavior strictly guaranteed across all installation methods. To lock down dependency versions, just update the `package.json` to specify exact versions of your project's dependencies (e.g., `"node-emoji": "1.4.3"`) rather than ranged versions (`"node-emoji": "^1.0.4"`), then adjust dependency versions in a controlled manner if/when needed (e.g., using `npm outdated`).

en/docs/_installations/alternatives.md
Daniel15: Hash verification is not the same. It doesn't verify **who** uploaded the
package or that it's actually a legitimate release, it's just a hash of the
contents. Additionally, npm uses SHA1, which is not considered secure.

Daniel15:
> You really should lock dependency versions in package.json

That won't lock transitive dependencies. Even if we depend on an exact version of package Foo (say, version 1.2.3), Foo could depend on `^1.0.0` of some other package (or even `*`, meaning **any** version).

Locking (including transitive dependencies) can be done with `npm shrinkwrap`, but [apparently it doesn't handle optional dependencies correctly](https://www.reddit.com/r/javascript/comments/5fyfqk/yarn_blog_lockfiles_should_be_committed_on_all/daodvyn/)

palmerj3:
@Daniel15 you should be able to do ```npm shrinkwrap --no-optionals``` to make a shrinkwrap that doesn't contain optional dependencies.

Daniel15:
But then I guess the problem would be that the optional dependencies don't
have their version numbers locked... I guess it's better than nothing.

palmerj3:
Optional deps, by definition, shouldn't be required in order for a project to function :)  If they are then that's a much bigger problem.

thejameskyle:
Yeah, but there's a difference in them not being there and them being there and being broken.

Daniel15:
@thejameskyle  - hopefully it's good now, let me know what you think 😃

PR #293, opened 2016-11-29
Add some notes about the safety of the tarball install method.

markstos:
- Note that it includes GPG signature verification.
 - For people rightfully concerned about piping downloaded code to bash,
   link to the source code of what's going to be run for inspection.

### DIFF
--- a/en/docs/_installations/tarball.md
+++ b/en/docs/_installations/tarball.md
@@ -8,8 +8,8 @@ your terminal:
 curl -o- -L https://yarnpkg.com/install.sh | bash
 ```
 
-The installation process includes verifying a GPG signature. The [source
-code for install script](https://github.com/yarnpkg/website/blob/master/install.sh) is available to review on Github.
+The installation process includes verifying a GPG signature.
+[View the source on GitHub](https://github.com/yarnpkg/website/blob/master/install.sh)
 
 You can also specify a version by running the following code in your terminal:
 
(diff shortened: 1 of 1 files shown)


thejameskyle:
Can you please check ["Allow edits from maintainers"](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

Daniel15:
Good idea 👍 

Is it worth linking to Github explicitly though? A user should download the install script from https://yarnpkg.com/install.sh to review its contents. I don't think we need the Github link.

markstos:
@thejameskyle I checked "Allow edits from maintainers"

@Daniel15 re: linking to Github: It's true that the source of `install.sh` could vary from the website to what's stored on Github. For someone concerned about security, it would also be a good practice to first download and then inspect code before running it, regardless of what someone linked to as what they claimed was the source code. I'm OK with removing the link to the source if it doesn't seem to add value.

Daniel15:
> For someone concerned about security, it would also be a good practice to first download and then inspect code before running it, regardless of what someone linked to as what they claimed was the source code. 

Yeah, that's why I don't think it's worth linking to GitHub. Anyone concerned about security would look at the source of the the script that's served from the website rather than the GitHub source, as there's no guarantee that they're the same :)

thejameskyle:
I think it's valuable just for getting people to contribute back in case they want to improve something.

PR #296, opened 2016-11-29
Fix #294: Validate  value provided to --version before using it.

### LINKED ISSUE #294: Install script needs better handling of invalid version numbers
To repro:
```
curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version lolwut
```

It should throw a proper error message when the download URL results in a 404 error.

markstos:
I manually tested both the success and failure cases. 

This patch uses the `--perl-regexp` option to `grep`. I did not attempt a comprehensive check of which OS versions support this flag. It is supported with the GNU grep that ships with Ubuntu 16.04.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,7 +17,7 @@ yarn_get_tarball() {
     # Validate that the version matches MAJOR.MINOR.PATCH to avoid garbage-in/garbage-out behavior
     version=$2
     if echo $version | grep -qP "^\d+\.\d+\.\d+$"; then
-      url="https://yarnpkg.com/downloads/$version/yarn-$version.tar.gz"
+      url="https://yarnpkg.com/downloads/$version/yarn-v$version.tar.gz"
     else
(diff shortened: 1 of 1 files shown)


Daniel15:
Perhaps we should just check for a download error rather than using a regex? `--version 99.99.99` will pass the regex but obviously isn't a valid version number, and we need to handle that too :)

cURL has a `--fail` flag that causes a non-zero error code to be thrown - The script could catch that and output a "Version number is invalid" error message?

markstos:
@Daniel15, I've now added the use of the `--fail` flag, so now we get a best diagnostic experience: An explicit message if the version format is bad and another explicit message if the URL fails to download.

Testing the script found a real bug: the correct use of `--version` was failing because we weren't redirecting to the correct URLs to download. I submitted an additional patch to fix that issue as well.

Daniel15:
Nice.

PR #298, opened 2016-12-01
Replace Perl regular expression with BRE

mythsunwind:
`-P` option for grep is not available on BSD which breaks the install.sh script on macOS when using the `--version` switch

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,7 +16,7 @@ yarn_get_tarball() {
   elif [ "$1" = '--version' ]; then
     # Validate that the version matches MAJOR.MINOR.PATCH to avoid garbage-in/garbage-out behavior
     version=$2
-    if echo $version | grep -qP "^\d+\.\d+\.\d+$"; then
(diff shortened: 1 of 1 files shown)


thejameskyle:
cc @Daniel15

Daniel15:
Seems reasonable - @markstos Since you added this validation in #296, what do you think?

markstos:
This PR did not work for me on Ubuntu Linux 16.04, but I just submitted PR #300, which can be tested on the Mac. PR #300 uses Extended Regular Expression (ERE) which are more common and compatible than the Perl Regex's I used before.  I tested that it correctly worked with these values:

     1.2.3 -> succeed
     99.99.99 -> passes regex, but receives "failed to download" diagnostic
     BOOM -> triggers "Version number must match MAJOR.MINOR.PATCH."

mythsunwind:
Will close this pull request, because of new PR #300

PR #300, opened 2016-12-01
Portability: Use more common "Extended Regex" instead of "Perl Regex"

markstos:
This fixes `yarn install --version` on with BSD grep, used by macOS.

This pull request is an alternative to PR #298, which did not work on Ubuntu Linux.


Daniel15:
Thanks! @mythsunwind Could you please test this one and see if it works in your environment?

mythsunwind:
@Daniel15  Works like a charm on latest macOS 10.12 Sierra

Daniel15:
Thanks for confirming!

PR #302, opened 2016-12-05
Fix #217: de-bower the docs

### LINKED ISSUE #217: Remove any docs on bower

markstos:


_data/i18n/en.yml
```diff
@@ -63,9 +63,9 @@ homepage_featurette_network_performance_title: Network Performance
 homepage_featurette_network_performance_description: |
   Yarn efficiently queues up requests and avoids request waterfalls in order
   to maximize network utilization.
-homepage_featurette_npm_and_bower_title: Multiple Registries
-homepage_featurette_npm_and_bower_description: |
-  Install any package from either npm or Bower and keep your package workflow
+homepage_featurette_npm_title: Multiple Registries
```
thejameskyle: Should probably rename this

markstos: Should we just drop the whole "Multiple Registries" section? Without mentioning Bower, the section heading "Multiple Registries" doesn't make much sense:

> ### Multiple Registries
> Install any package from npm and keep your package workflow the same.

thejameskyle: If we just change the title to "All the same packages" or something like that it would be fine

markstos: I updated the wording to be "Same Packages". All the other section heads were just one or two words and capitalized the words, so this revision should fit in that respect. 

You can check how it looks on Netlify: https://deploy-preview-302--yarnpkg.netlify.com/

Issue #306, opened 2016-12-07
No `yarn run` shortcuts documented


mgol:
Yarn, similarly to npm, has a few shortcuts for `yarn run`, e.g. `yarn start` cane be used instead of `yarn run start`. They should be documented.

One thing that surprised be was that `yarn build` works as an alias to `yarn run build` which is different to what npm does so I can't just rely on the npm docs for that.

PR #307, opened 2016-12-07
Document the default value of the `start` script.

markstos:
This addresses part of the issues raised in #306.


markstos:
@Daniel15 I've amended my comment to move the docs for this default value based on your feedback.

Daniel15:
Thanks!

PR #312, opened 2016-12-14
Fixes yarnpkg/yarn#2122 - mktemp invalid argument

matteocng:
[BusyBox](https://busybox.net/BusyBox.html)'s mktemp requires that:
> TEMPLATE (-t parameter) must end with XXXXXX (e.g. [/dir/]nameXXXXXX).


Daniel15:
Thanks!

PR #316, opened 2016-12-20
docs(search): only init docsearch when relevant

vvo:
Previously we were instantiating docsearch on every page resulting in
console errors:

2016-12-20-184441_1444x763_scrot

Now fixed


vvo:
Need small change to avoid loading scripts in the middle of the page

vvo:
PR updated and ready

PR #317, opened 2016-12-21
docs(search): only init docsearch when relevant

vvo:
Previously we were instantiating docsearch on every page resulting in
console errors.

2016-12-20-184441_1444x763_scrot


Daniel15:
Thanks!

PR #321, opened 2016-12-21
Fix language dropdown

### LINKED ISSUE #310: Language switcher does nothing when clicked
When I click on the header menu item that says "English", I expect some kind of drop-down menu or language picker to appear. Nothing is currently happening for me, either in Firefox or Chrome.

thejameskyle:
Fixes #310


Daniel15:
Nice, thanks!

PR #331, opened 2016-12-31
Add title for /search.html

vjeux:
image


Daniel15:
Thanks! I think I missed this while reviewing because I have so many Chrome tabs open that I can't actually see the titles 😛  We should probably update the title whenever a search is performed, too.

Haroenv:
This did set `og:title`, but not the `Package search | yarn`, because the `default.html` checks if layout == post. Should that logic be changed?

https://github.com/yarnpkg/website/blob/e8f2373cb6265eb3fe10775b92bca66694cf483a/_layouts/default.html#L15

PR #332, opened 2017-01-02
fix(search): url pattern changed on NPM

bobylito:
Hi :) We noticed a change in the URL pattern on NPM. Right now it's mostly invisible since they have a redirect, but it doesn't work with all the packages (`search` for example). Fixing it to make it more future proof :)


Daniel15:
Thanks!

PR #334, opened 2017-01-04
Rename to list.md and change "ls" to "list"

Ash-v-G:
current version v0.18.1 links to https://yarnpkg.com/en/docs/cli/list when you run ls command, which is a 404.

the documentation needs to reflect the change from ls to list.


Haroenv:
I think this is the same as #333

bestander:
Looks like it is a duplicate

PR #335, opened 2017-01-04
add a page title to the /search page

Haroenv:
it wasn't showing up before, because the layout file didn't have a case for a non-blogpost that had a title, so now there's another condition for that.

see #331 where a page.title was added to that page


Daniel15:
This seems like a safe website change to me.

Haroenv:
Top!

PR #336, opened 2017-01-05
ci: Update yarn cache directory

mc10:
yarn is [now cached in ~/.cache/yarn](https://github.com/yarnpkg/yarn/blob/a1192e596b85acb2f63daa1cbf30e662c2125310/src/constants.js#L41), not ~/.yarn-cache.


Daniel15:
Thanks!

PR #337, opened 2017-01-06
Remove Bashism from install script

mattclegg:
Currently the install script depends on bash brace expansions over strings. Although this syntax is useful for abbreviation it is not portable. Removing this brace expansions allows the script to be run in alpine linux.

PR Tested using;
```
FROM mhart/alpine-node:6.9.3

RUN apk add --no-cache --virtual .yarn-install-deps \
        curl \
        gnupg \
        tar

COPY install.sh /install.sh
RUN /install.sh
ENV PATH="/root/.yarn/bin:${PATH}"

CMD ["yarn"]
```


bestander:
cc @Daniel15

install.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
Daniel15: One of the reasons we explicitly used bash was because we couldn't get the colour codes working correctly in Dash, which Ubuntu and Debian use by default. I tried tput a while back, but that caused other issues. Could you please get the colour codes working too?

install.sh
```diff
@@ -25,19 +25,30 @@ yarn_get_tarball() {
...
-    mkdir .yarn
-    tar zxf $tarball_tmp -C .yarn --strip 1 # extract tarball
-    rm $tarball_tmp{,.asc}
-  else
-    printf "$red> Failed to download $url.$reset\n"
-    exit 1;
-  fi
+  # Get both the tarball and its GPG signature
+  download_to_tmp $tarball_tmp $url
+  download_to_tmp $asc_tmp $url.asc
```
Daniel15: This will download the two files sequentially, whereas the old code downloaded them in parallel. Downloading sequentially will unnecessarily slow down the installation process by adding another round-trip. Could you please do it in a way that still downloads the files in parallel?

install.sh
```diff
@@ -25,19 +25,30 @@ yarn_get_tarball() {
...
+    if curl --fail -L -o "$1" "$2"; then
+      printf "$green> Downloaded $2.$reset\n"
+    else
+      printf "$red> Failed to download $2.$reset\n"
+      exit 1;
+    fi
+  }
+
   tarball_tmp=`mktemp -t yarn.tar.gz.XXXXXXXXXX`
-  if curl --fail -L -o "$tarball_tmp#1" "$url{,.asc}"; then
```
Daniel15: Maybe you could just do this:

```
if curl --fail -L -o "$tarball_tmp#1" "$url" "$url.asc"; then
```
?

mattclegg: Yeah I tried it that was first of all, but the problem is that the second file doesn't get saved to the temp file. See;
```
/ # curl --fail -L -o "$tarball_tmp#1" "$url" "$url.asc"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    92  100    92    0     0     30      0  0:00:03  0:00:02  0:00:01   190
100   595    0   595    0     0     98      0 --:--:--  0:00:06 --:--:--  1126
100 3537k  100 3537k    0     0   285k      0  0:00:12  0:00:12 --:--:--  915k
-----BEGIN PGP SIGNATURE-----

iQIzBAABCAAdFiEEagEMUWYAZZmqF/CBRsITDf0kl/UFAlhVL+QACgkQRsITDf0k
l/V1WA/+LkAyUUaW3hYLfrznPtqNOcnQGcgdYBqPPgNkZ5teHk9/Xjn1etpoA+Za
1aMETfPLkVxm9wlMrRERtQvWNmLm3cnxCrrPAAvoGQMw9nfNqkbDUWWVeYZwb3R8
C9Hcl/CSvl0rfl1rcBL6YCSkdguo92L1ENjXFLTaNx/0sXhNcvSeRV9PaP1xlJde
Vp+/SAs/Wh60H2oqGRyXN/8e7qhbCpuxsVXrUHfzcZJ4hNfDhnZ3/YRSGn29MTZa
lpDU53zpvwgQT8NpcMLdgafCuNnocTrcMdtOaoxvfTs9cROVu3e5lQh1u9ALFLu4
VJRO8rVVoyTzh7NyFOJT1NxLgFMji/5yUemm92dAEN1fIzG6ZGrEk8Q4H7o0XsoR
CkOc275X5d5JWmmBn0Nitk5FX9RLzYVCTGkKjg5cHZeF1wu0ouOwQ1kYhtBPzB3s
LXGbH9ZMRV6UgZZHo9rIz0V7OyQv3tzQJPMUyWNXnysN0LYDY6+wXv3wI22f0yr9
YHrI0r4oEjrpQzake54okQl4hon+sdQ1iUd/jkN52mz3CFoqWwlqq02iHt19WGs0
yFRmfI6glOYew9DWZOADejITvOgDRlSTElju3nSvLkkzs9Hd3AX43aVodJonPfBT
YE8FvWUPYwUSirz/ABLTDcVAz1G4wXzWVE4M9E7DxVmM+JJTJIc=
=5/GI
-----END PGP SIGNATURE-----/
```

I decided on having it split as a function for readability would make more sense?

install.sh
mattclegg: I think the downloads are actually happening asynchronously so splitting it to 2 different sequential requests shouldn't make a difference.

install.sh
mattclegg: Ok, updated.

Daniel15:
Thinking about this further, I think cURL itself is doing the expansion, rather than Bash? Running `echo "$url{,.asc}"` in Bash just gives me that value verbatim (eg. `https://yarnpkg.com/latest.tar.gz{,.asc}`). Furthermore, I just tried this in the version of Dash that comes with Ubuntu 14.04 and it worked as expected:

```
curl --fail -L -o '/tmp/yarn337test.tar.gz#1' 'https://yarnpkg.com/latest.tar.gz{,.asc}'
ls /tmp/yarn337test*
```

Daniel15:
Just had a look at the code, and it looks like cURL does indeed do the globbing itself: https://github.com/curl/curl/blob/ee3c83f39c90126cabb9be896931725c32f22e09/src/tool_operate.c#L450-L453

So in theory the old code should work on other shells such as Dash. The changes for the colour codes would be great to merge, though! 😃

mattclegg:
Although cURL does handle the globbing the script it still relying on bash brace expansions to remove the temporary files. When running the current script in `alpine-linux` you receive the error;
```
rm: can't remove '/tmp/yarn.tar.gz.XXXXHhAgle{,.asc}': No such file or directory
```

This is originally what sent me down the line of refactoring the implementation of cURL. Or would you rather just have [this line](https://github.com/yarnpkg/website/blob/master/install.sh#L36) changed to `rm /tmp/yarn.tar.gz*`?

Daniel15:
Yeah, simply updating the rm command should be fine :)

mattclegg:
Updated as requested.

mattclegg:
@Daniel15 Anything else need changing on this?

Daniel15:
Thanks for updating the PR, looks good now!

Issue #339, opened 2017-01-07
Update "self-update" page


tobymurray:
Given this issue [here](https://github.com/yarnpkg/yarn/issues/1139) and in particular [this comment](https://github.com/yarnpkg/yarn/issues/1139#issuecomment-268229087) it'd be nice to have some kind of hint that `self-update` is gone until further notice. Not being aware of its removal, it was not trivial to track down what was going wrong (I assumed I had botched something up myself), and it's actively misleading that the website currently shows the presence of the command - https://yarnpkg.com/en/docs/cli/self-update.

Even a "note, this is in active development, see <github issue>" or something providing any information would go a long way.

tobymurray:
Fixed by [PR#340](https://github.com/yarnpkg/website/pull/340)

PR #340, opened 2017-01-07
#339 Add note explaining no self-update in 0.18.1

tobymurray:
The self-update feature was removed pending a more robust
implementation. At present, it's not clear from the CLI or otherwise
that self-update was removed. This gives an indication in the
documentation that at least the user is not using it incorrectly.

### DIFF
--- a/en/docs/cli/self-update.md
+++ b/en/docs/cli/self-update.md
@@ -8,6 +8,6 @@ layout: guide
 
 ##### `yarn self-update` <a class="toc" id="toc-yarn-self-update" href="#toc-yarn-self-update"></a>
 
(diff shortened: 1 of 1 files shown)


en/docs/cli/self-update.md
```diff
@@ -8,4 +8,6 @@ layout: guide
 
 ##### `yarn self-update` <a class="toc" id="toc-yarn-self-update" href="#toc-yarn-self-update"></a>
 
+*Note: `self-update` is not available in version 0.18.1, which means updates must be done manually. See <a href="https://github.com/yarnpkg/yarn/issues/1139">issue #1139</a> for details.*
```
Daniel15: Maybe just omit the version number and say "self-update is currently not available"

tobymurray: Works for me - did so in 805ea31

PR #341, opened 2017-01-08
feat(search): Also highlight description

vvo:
Description now shows the query matches in an highlighted fashion like
others fields are.

**Before**
image

**After**
image


Daniel15:
Thanks!

PR #348, opened 2017-01-16
Docs: Consistent use of Upgrading / upgrade

Undistraction:
Given that the command is called `upgrade`, the title should echo this.


Daniel15:
Good catch! Thank you 😃

PR #356, opened 2017-01-20
Update `yarn list` command docs

grant:
The `ls` documentation is currently incorrect.
`yarn ls` -> `yarn list`

There could be more effort put into this and in the yarn package itself (i.e. `yarn global ls` is still the norm).


Haroenv:
related to #333 and #334

PR #364, opened 2017-01-31
Prepare site for translation

thejameskyle:


index.html
```diff
@@ -1,4 +1,13 @@
...
+layout: null
 ---
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title></title>
+  </head>
+  <body>
+    <a href="lang/en/">Click here to see the site</a>
```
Daniel15: I assume this is only visible when running the site locally? Could it auto-redirect (eg. via a meta refresh tag)?

Issue #367, opened 2017-02-01
Website links return 404


Ionaru:
On https://yarnpkg.com the 'getting started' menu button links to https://yarnpkg.com/docs/getting-started, this link goes to a 404 page. The correct link should be https://yarnpkg.com/lang/en/docs/getting-started/

Seems the language choice is not preserved while linking through the website, which results in a wrong URL.

Using Opera 42.0.2393.517 on Ubuntu 16.10
Browser ID: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.87 Safari/537.36 OPR/42.0.2393.517

DomClayton:
Also possibly related, when visiting https://yarnpkg.com/ I'm seeing a page with just this link: 

> [Click here to see the site](https://yarnpkg.com/lang/en/)

I'm expecting this is meant to automatically redirect.

onemenzel:
I get this using Opera and Safari in their newest versions. Firefox and Chrome are working though.

amitmtrn:
clicking on getting started link to https://yarnpkg.com/docs/getting-started
and should link to https://yarnpkg.com/lang/en/docs/getting-started/

israphim:
workaround: click on the language selector and manually select english

Daniel15:
Does anyone in this thread have their preferred language set to something other than English? My hunch is  that there's a bug that's occurring for anyone that uses a language that the site doesn't support.

Daniel15:
cc @thejameskyle

onemenzel:
German here. Now that you mention it: my default language is set to English in Firefox. And on my language list in Chrome it's No. 1.

Daniel15:
Thanks for the confirmation. I can repro in Firefox if I change my language preference to German:
![](http://ss.dan.cx/2017/02/firefox_01-15.17.40.png)

gotofritz:
Doesn't work for me on either firefox or chrome canary, but it works on safari.
I am in Germany with an English OS and browsers.

thejameskyle:
I figured out the issue, I'll publish a fix soon

thejameskyle:
Fixed

petermikitsh:
I still can't access the getting started page. I wanted to setup yarn on my CentOS CI server :(

PR #368, opened 2017-02-01
added auto-redirect

GiacomoLaw:
This will automatically redirect the person to the site. When there are more translated versions, simply remove the script from the header.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,8 +6,11 @@ layout: null
   <head>
     <meta charset="utf-8">
     <title></title>
+    <script>
+      window.location = "https://yarnpkg.com/lang/en/";
+    </script>
   </head>
   <body>
-    <a href="lang/en/">Click here to see the site</a>
+   <!-- <a href="lang/en/">Click here to see the site</a> -->
   </body>
 </html>


amitmtrn:
I think it would be better to use [jekyll redirect from](https://github.com/jekyll/jekyll-redirect-from) plugin
```
---
layout: null
redirect_to: lang/en/
---
```

Daniel15:
Redirects should go in the _redirects file in the root so they can be done server side. 

Having said that, I don't think this behaviour is expected, I think it's supposed to automatically default to English. I think there's a bug elsewhere.

amitmtrn:
@Daniel15 I can't find how the server redirects should work on github pages
https://help.github.com/articles/redirects-on-github-pages/
I understand that `_redirects` is responsible for the server redirects but from what I recall server redirection in jekyll should go through `.htaccess`, I'm not familiar with the `_redirect` file, can you refer me to documentation regarding this?

Daniel15:
We're using Netlify, not Github Pages. They have basic features that Github
Pages lacks, including redirects. :)

https://www.netlify.com/docs/redirects/

GiacomoLaw:
Ah, okay. But this is just a simple JS script in the header.

Daniel15:
This is going to be fixed as part of #367, so we won't need the redirect.

GiacomoLaw:
Okay, cool!

PR #369, opened 2017-02-02
fix typo in run.md

jdanyow:


jdanyow:
I don't have access to the netlify build log- let me know if anything needs changing- guessing there was a timeout or something.

Daniel15:
I think the old version is correct, as earlier in the page it defines the "test" script as:
```
"test": "jest"
```

PR #371, opened 2017-02-03
Fix install command for codeship

jakul:
Codeship use `nvm` to choose the correct node version based on `package.json`. Globally installing on Codeship uses the system node version (0.10), which Yarn is not compatible with.


Daniel15:
When using npm to install Yarn, it does need to be installed globally otherwise running `yarn` won't work. Can you upgrade the system Node version on Codeship?

jakul:
At the moment it's working OK for me with local install.

The system version on Codeship is stuck to 0.10. They suggest using nvm to
select newer versions, which works fine for everything else except for
globally installed packaged!

Em seg, 13 de fev de 2017 às 07:05, Daniel Lo Nigro <
notifications@github.com> escreveu:

PR #374, opened 2017-02-07
docs(search): fix empty query display

vvo:
Before this commit when rapidly typing a letter then removing it, the
"no query" screen (search homepage) was not showing because of a race
condition.

This is now fixed, I also upgraded to latest instantsearch since the
default url sync mechanism is now the good one.

Bad behavior:
bug-empty-query

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -11,7 +11,6 @@
     indexName: 'npm-search',
     urlSync: {
       trackedParameters: ['query', 'page', 'facetsRefinements'],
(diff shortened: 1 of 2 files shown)


Daniel15:
Looks like there's some merge conflicts here, most likely due to #375. Could you please rebase? Thank you!

vvo:
Sorry, given what @Haroenv is planning this is outdated and will ultimately be fixed with a new PR.

Daniel15:
Cool, thanks.

PR #375, opened 2017-02-08
Allow ES6 modular code via Babel and Webpack

### LINKED ISSUE #330: Add Babel support to site build
As mentioned in https://github.com/yarnpkg/website/pull/322 - add Babel to this site so we can use more ES6 features.
`make build` should build the JS files.
Need to ensure Netlify is using `make build` to run the build.

Haroenv:
the tooling now includes yarn itself (which can be relied on, since netlify has yarn).

closes #330

The structure of the JS files now is (in my opinion) more reasonable to think about, and also will allow for faster development for further features.

jQuery, bootstrap are still loaded via cdn, because that's better for large bundles like those.

instantsearch, docsearch, moment are loaded via npm and bundled in (which saved approx. 300kb on the search page)


Haroenv:
hmm, netlify seems to not run the build step. Could someone check out what steps are done so I can troubleshoot?

Makefile
```diff
@@ -37,17 +41,15 @@ crowdin-download: test-crowdin
...
 
-test-bundler:
+test-builddeps:
 ifndef BUNDLE_EXISTS
 	$(error bundler is not installed. Run `gem install bundler`)
 endif
-
-test-jekyll:
-ifndef JEKYLL_EXISTS
-	$(error Jekyll is not installed. Run `make install`)
```
Daniel15: Why remove this? The site still depends on Jekyll so the script should probably check that it exists.

_layouts/pages/docs.html
```diff
@@ -1,8 +1,7 @@
 ---
 layout: page
 scripts:
-  - https://cdn.jsdelivr.net/docsearch.js/1/docsearch.min.js
-  - /js/docsearch.js
+  - "/js/build/documentation.js"
```
Daniel15: Would be good for these file paths to have a cache breaker... Can we do that? You can use `assets-webpack-plugin` to write the cache breakers into a JSON file if you use a `filename` like `[name].js?[hash]`, could the site read that JSON file?

js/src/lib/deeplink.js
```diff
@@ -0,0 +1,30 @@
+export function handleTabs() {
```
Daniel15: What's the advantage of putting this in a separate file, given it's tightly-coupled to the page? I'd suggest just putting this in `js/src/install.js` directly

webpack.config.js
```diff
@@ -0,0 +1,34 @@
+const webpack = require('webpack');
+
+module.exports = {
+  entry: {
+    common: './js/src/common.js',
+    documentation: './js/src/documentation.js',
+    install: './js/src/install.js',
+    nightly: './js/src/nightly.js',
+    search: './js/src/search.js',
```
Daniel15: If you want to use Webpack, I think this should be done using code splitting... With this setup, some code required by two of the files (eg. if both `install.js` and `nightly.js` use the same module) will be duplicated across both bundles. Code splitting can pull the common code into a separate module that both can use.

Daniel15:
I like this idea 😃  I wonder if we could just use Babel for now, to reduce complexity? We don't really need any Webpack features yet.

webpack.config.js
Haroenv: totally right, will be fixed!

_layouts/pages/docs.html
Haroenv: I wanted to do that, but didn’t figure out how it’s done yet, your proposal sounds good.

Makefile
Haroenv: Because jekyll is executed via bundler, not via the jekyll standalone binary

Haroenv:
Thanks for your review @Daniel15, It’s evening now for me, so I’ll answer quickly for now. 

I’d prefer to have webpack in here, because then I can easily depend on let’s say React or something else to make the detail pages for the search. 🎉

vvo:
(I worked a bit with @Haroenv on this)

Using only babel may work I guess even if we want to use React, we would have to inject React via script tags then (because Babel won't be able to inline what's inside node_modules right?).

It would also mean that we would not be able for now to use any dependency from npm, this may be a bit tricky at some point to have to rely only on script tags. As you want!

js/src/lib/deeplink.js
Haroenv: I did this because I imagine that pages other than the install page could benefit from having the handling of tabs into the history

_layouts/pages/docs.html
Haroenv: This is done now @Daniel15

webpack.config.js
Haroenv: Hey @Daniel15, thanks for the suggestion. I've never done code splitting before, so now I manually did what code would be split too, with a hand-made common.js that hosts things that all pages have. Could you tell me how that would be set up properly? Thanks!

Haroenv:
I fixed some small issues that creeped in, but now it should be good to merge I think 😄

Daniel15:
This looks pretty good to me. Tested a few pages on the Netlify preview (https://deploy-preview-375--yarnpkg.netlify.com/) and it all seems to be working.

The only thing I don't like is that there's two separate "build" commands: The build command in the make file (`make build-production`), and the build command in package.json (`yarn run build`), with no clear documentation on why this is the case. To me it seems like you could completely avoid the `build` script in `package.json` and just call `node_modules/.bin/webpack` in the `Makefile`. I'm going to merge this as-is, but that's an idea for cleaup if you submit future diffs. Thanks for your contribution!

PR #383, opened 2017-02-17
Packages search: allow searching for packages on every pages

Haroenv:
> We want to allow people to search for packages no matter the page they are in, instead of having to click on "Packages" in the header.

>Adding the package page will require a fair amount of frontend code and templating. We propose to add babel and React to the yarn website stack and reorganize some already present code in a module way.

To do this, we included React in the process, and used [react-instantsearch](https://community.algolia.com/instantsearch.js/react/). 

Searching works like this now: 

* you are on a page (for example /docs/getting-started)
* you search something in the search bar
* the url becomes /packages?q=something
* you click the `X` or erase your search
* you're now back on your original page (/docs/getting-started)

If you'd refresh the page or share the url, erasing the search would get you to the featured package page, which makes sense. 

also improves the webpack build time from sometimes quite long to about a second now.

I noticed that the style we had for docsearch (in /docs) doesn't actually apply, and when I changed the order of the css links, I saw that the custom style didn't really work, so I put it to the default on purpose.


Haroenv:
There's no internationalisation in the search bar queue, but I didn't know how to get to the site.data.i18n from JS

vvo:
We could inline that data in the page as a an inline script tag in templates, or have a build step that creates a Js file out of it.

vvo:
Added:
- separated vendor bundle with jquery-slim, bootstrap selected plugins, react, react-instantsearch (cc @Daniel15)

_includes/i18n-as-script.html
```diff
@@ -0,0 +1,11 @@
+{% comment %} get the current translations from Jekyll, to be used in scripts {% endcomment %}
+<script>
+  window.i18n = {
+    url_base: '{{url_base}}',
+    active_language: '{{vars_active_language.tag}}',
+    search_placeholder: '{{ i18n.search_placeholder }}',
+    search_by_algolia: '{{ i18n.search_by_algolia }}',
+    search_by_read_more: '{{ i18n.search_by_read_more }}',
```
Daniel15: I suspect you should be JSON encoding these, as the translations in some languages may have apostrophes in them. We can follow up on that separately.

_sass/_base.scss
```diff
@@ -19,3 +19,31 @@ svg {
...
...h63dZS5/aMszqeuUwn840yvmYvW3KphDlJ46lvLnMxBi65MH6XEw4mrTqPGhS+0pbaONZr/3PctiR9maNT53dR05cZzVnqPLScB86dtThqmTeUeUed8t04oOd1Dp3nzmCHZ+5/icLnGa+17bRjblDik3lZHr8eN3ObLkl/c0AXxwjm7GkK9f4ywfj8hMJq3EtQuLcRrNOn5dRMMmqRoG9PSs8u3W9SWMR7KOzL8wTTc5xgmrqt6d7o2Szwa/V5m0iQOel6VG0eIhJiFniXwjC13hngTfUj+daisFNzvctOAbmHYHBOAi+q/yKFru/2FwkW+TmCcn5OY7cp7Fpj5SDPJlwfIGJhm0YpC+4OonzhYYJ1arr/qMbdLn9mCWb0HuB9jXmRYL2aBf4gwfJ+UbbNESzZt+SPY9yJ0TwE/JSrWdDHpbdJxH9A49+lMZ7Ts7cJVvodwO+TjtMEW3tRffZq3Pcp5TtDrGZ8/0mfp+TvNoLJb9b4u7L720R+LyUf5ygTqkFhp9dZ6weJObJfbXO8KyJeLncxi7jSPZervEzEH/lwTP6syO7dlNIcl/ycJGJrO82gfopSYgNlQa6I+LeSDV40W8Ix2+Q6tXdkywRRSvOR8H5W/riawS8IL4QN9TXJ+ARRUXFSNj9KeWnNAD8gykaO6PlzwAsDwK9QXQYRzPeIYK1Vz2EQtwK/IWpVpih072Fikrt+5IAcXNbg71OKGLcJ2MMEhXtCf/cRE85v0MsUSv1hItiH5eC9suVxoh7tb7LxY0odDXK+IR259iazctvStaDxpgTca3o+S0wuM6jtQ702yJPwDLEQma5/nEi0emGvi09H1cdvwTEiQT+l1Df9jtjBDFHeUF4Y5il1TjuF96vC/7hsuEIsVjvUZ1E4TQL/JGrNXIDYSdpq9wiRN17UTkq/y0BcZzYlLD6SvZeJiX6MiE2bUph6Wv0XiZeby13szw7Z9pLanxKWrot7Qu0nKLVOy8J2msjtP1MmeWZq36HPhXTfi3tFTLZJfebav1OyO++AKuHiOsIWMQ8WiLys5Nug7r2pP4+PMDqr6xEiZhVR67hftn2QbH9A7euseoj8cn3lEUrd5SXgH2pzJ6V85x4ivjME5q9I535hvULUo50n8vNf6vuG+h+RTYOyc7vaviTsvCYc+i9pQJWwuXILrwAAAABJRU5ErkJggg==);
```
Daniel15: Does SASS have a way to inline images like this? I feel like it'd be more maintainable if you had the images in external files, and used a SASS function to inline them.

_includes/i18n-as-script.html
Haroenv: Great point, it was encoded earlier but seemed redundant to me.

Daniel15:
This looks good to me! Let's get this in and iterate as needed :)

_includes/i18n-as-script.html
Daniel15: @Haroenv - It also prevents someone from performing XSS via translations (eg. if someone were to somehow enter the translation as `foo'; window.location = 'http://badsite.com/';`) 😃

Haroenv: all right, I updated it just before you merged, so we’re safe for that now

PR #388, opened 2017-02-22
replace ./.yarn -> $HOME/.yarn

trevj:
This causes the install script to fail unless the current directory is `$HOME`. This is not the case if, like me, you are running the install script as part of building a Docker image for a continuous build.

Note that all other references to `.yarn` are **already** `$HOME/.yarn`


Daniel15:
Thanks!

PR #389, opened 2017-02-25
check for writable files

umeboshi2:
I am using [fresh](https://github.com/freshshell/fresh) to manage dotfiles.  This patch will make sure the dotfile is writable before attempting to modify, and will exit with instructions for adding yarn to path if script cannot modify dotfile.  I have only tested this for .bashrc on my laptop.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -109,7 +109,7 @@ yarn_link() {
 }
 
 yarn_detect_profile() {
-  if [ -n "${PROFILE}" ] && [ -f "${PROFILE}" ]; then
+  if [ -n "${PROFILE}" ] && [ -w "${PROFILE}" ]; then
     echo "${PROFILE}"
     return
   fi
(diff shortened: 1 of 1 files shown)


bestander:
@Daniel15 what do you think?

badersur:
I think yarn should check if write permission is granted to the profile files but I think the message should be changed a little bit (To say it doesn't have write permission instead of profile not found when it's available!)

and
```sh
if [ -z "${YARN_PROFILE-}" ] ; then
    printf "$red> Profile not found. Tried ${YARN_PROFILE} (as defined in \$PROFILE), ~/.bashrc, ~/.bash_profile, ~/.zshrc, and ~/.profile.\n"
    echo "> Create one of them and run this script again"
```

doesn't seem alright to me because the `if` checks if the variable is empty then it tries to print it! Will discuss this in [my other PR](https://github.com/yarnpkg/website/pull/543) where we work on using `sh` syntax. (not sure & I don't know what `-` is doing in  `${YARN_PROFILE-}`...)

PR #394, opened 2017-03-01
add version flag to travis container-based builds

le6o:
The previous section of the page, "sudo-enabled builds", recommends locking the version of yarn when installing for ci. (See the deb-specific-version.md include)

No similar example is given for `container-based builds` so this simply updates that example.


Daniel15:
Thanks!

PR #395, opened 2017-03-01
feat(documentation): i18n on the docsearch

Haroenv:
The documentation has been written in a few languages now, but was only crawled in English. 

Now the setup has been put in place for documentation in multiple languages (see https://github.com/algolia/docsearch-configs/pull/130 as well). 

The other languages can only be added as soon as this is merged.


Daniel15:
What happens if the language is not indexed in Algolia? Does it fall back to English, or will no results be returned?

Haroenv:
If the language is not yet indexed it will give English results

PR #396, opened 2017-03-03
Update install.md

VikiAnn:
I was trying to understand the problem described in PR #2782 and I noticed this flag is not documented.


Daniel15:
Nice! I didn't realise this existed.

PR #397, opened 2017-03-03
feat(docsearch): only use active language

Haroenv:
so the syntax changed a little as to what I expected, and this PR is needed to make sure that only the current language is shown in the result.

If the language isn't in the list of languages indexed by algolia ([docsearch-configs](https://github.com/algolia/docsearch-configs/blob/master/configs/yarnpkg.json))

cc @daniel15


Haroenv:
Without this patch searching will show all languages, instead of the current language. Could this be merged soon @Daniel15?

Daniel15:
Thanks! Sorry for the delay.

Haroenv:
Great! Sorry for confusing the options previously

PR #401, opened 2017-03-04
feat(docsearch) mark placeholder for translation

Haroenv:
Since the documentation is available in multiple langauges, the prompt should be too


Daniel15:
Good catch!

Haroenv:
For some reason this doesn't show up in other languages, I don't see why that would be the case, I thought it fell back on English if a string isn't translated yet?

PR #409, opened 2017-03-08
Add some offline mirror docs

dguo:
Docs for yarnpkg/yarn#2836. @bestander said we can convert the [blog post](https://yarnpkg.com/blog/2016/11/24/offline-mirror/) later.

### DIFF
--- a/lang/en/docs/prune-offline-mirror.md
+++ b/lang/en/docs/prune-offline-mirror.md
@@ -1,0 +1,22 @@
+---
+id: docs_prune_offline_mirror
+guide: docs_offline_mirror
+layout: guide
+---
+
+After you configure your offline mirror, Yarn will automatically add new
+package tarballs to the mirror. However, it does not automatically remove
+tarballs that are no longer referenced in `yarn.lock`. For example, if you `$
+yarn remove` a dependency, the tarball will remain in the mirror, even if no
+other dependencies have it has a sub-dependency. This behavior can be desirable
+in a setting where many projects share the same mirror, but when that is not
+the case, you may want to have Yarn remove unnecessary tarballs.
+
+To turn on automating pruning, set `yarn-offline-mirror-pruning` to `true` in
+your `.yarnrc`:
+
+`$ yarn config set yarn-offline-mirror-pruning true`
+
+Now, tarballs will be removed when appropriate. The end result is that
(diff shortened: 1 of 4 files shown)


lang/en/docs/prune-offline-mirror.md
```diff
@@ -0,0 +1,22 @@
...
+yarn remove` a dependency, the tarball will remain in the mirror, even if no
+other dependencies have it has a sub-dependency. This behavior can be desirable
+in a setting where many projects share the same mirror, but when that is not
+the case, you may want to have Yarn remove unnecessary tarballs.
+
+To turn on automating pruning, set `yarn-offline-mirror-pruning` to `true` in
+your `.yarnrc`:
+
+`$ yarn config set yarn-offline-mirror-pruning true`
+
```
rmacklin: Is it worth mentioning the version this feature was added in?

Actually, now that I look, there isn't yet a release or prerelease containing this, so having it live on the docs page is sort of confusing. But once there is a release, I think it could help to mention that this feature was introduced in that version.

The reason is that people using this feature would probably want to ensure everyone on their team is using at least that version, otherwise a team member with an old version of yarn could modify dependencies and the pruning wouldn't occur, which would then cause unrelated pruning for the next developer who modifies dependencies and is using a version that supports pruning (which is potentially confusing/annoying).

Personally I get around version mismatches like this by committing the bundled yarn executable into the repository and ensuring that team members are invoking that version (e.g. running `bin/yarn` rather than `yarn`). However, I know that many don't do this, and regardless I think it'd still be useful to mention the version that added pruning (once there's a release).

dguo: Yeah, this is a good point. It didn't occur to me that the Yarn docs aren't versioned along with Yarn itself. @bestander, how do you want to handle this? We could add a message now saying that it will be in the next Yarn release, and then change it to the actual version when it happens.

bestander: Currently we don't version the website, over time we will probably do it, the same way we do for React Native.

In our team we check in Yarn js bundle where it is important for all the team members to be on the same version.

I don't think it is a big deal for this feature as anyone changing dependencies after the team memebers with old Yarn would clean everything up.

PR #406, opened 2017-03-07
feat(search) add opensearch.xml

Haroenv:
closes #405

All browsers that support the "search in this site" spec use this version.

Issue #405, opened 2017-03-07
Package search as search provider in Firefox


esphen:
Hi,

I love the new package search on the yarn website. I love it so much, in fact, that I would love to have it as a search provider in the search bar. It would be amazing to just be able to type `n react`. in the search bar and have it open to the yarn package search page with the term filled in and the search bar focused.

Usually in Firefox, you are able to add search providers when you [access the website][add]. This is not an option on the yarn website, as far as I can see. What sort of heuristics does Firefox use to detect search fields? Is this something yarn can mimic and add to the website?

[add]: https://support.mozilla.org/t5/Manage-preferences-and-add-ons/Add-or-remove-a-search-engine-in-Firefox/ta-p/34220

vvo:
Dunno if related but in google chrome I am able to get a GitHub and npm search:

![image](https://cloud.githubusercontent.com/assets/123822/23658109/87a595d8-0341-11e7-8d19-b2dfbc0653e1.png)
![image](https://cloud.githubusercontent.com/assets/123822/23658111/890da316-0341-11e7-9eda-10727bc36dd3.png)

But does not work for yarnpkg.com.

You question seems to be a different one, we would need more firefox documentation on what you want to achieve to do it.

esphen:
A quick search turned this up. Looks like what's needed to make it work with yarnpkg.com:

https://developer.mozilla.org/en-US/Add-ons/Creating_OpenSearch_plugins_for_Firefox#Autodiscovery_of_search_plugins

PR #410, opened 2017-03-08
fix(opensearch) link to /en/ by default

Haroenv:
because a redirect takes you from /packages?q=whatever to /en/packages, which forgets the querystring, linking to en by default (from which you can then change your language) seems like a good solution.

see also #406 and #405


Daniel15:
> because a redirect takes you from /packages?q=whatever to /en/packages, which forgets the querystring, 

I wonder if this is a Netlify bug, or if there's some option we need to set (akin to QSA / Query String Append in Apache)

Haroenv:
Yes, this might be fixable in the _redirects too, but I didn't see an obvious way

PR #412, opened 2017-03-08
fix(search): Fix OpenSearch file in Firefox (#405)

esphen:
Firefox does not show the ability to add the site as a search provider when there is an xml declaration at the top of the file.

This commit removes it so that the search works with Firefox.

Also add a title to the link, as suggested by the MDN docs

Confirmed to work on localhost:4000

opensearch_ff


Haroenv:
Cool! Thanks for helping out

PR #414, opened 2017-03-09
fix(url) handle deep linked tabs before handling the current OS

### LINKED ISSUE #413: Not possible to link to a specific OS tab on the install page
**STR:**
1. Visit https://yarnpkg.com/en/docs/install
2. Use the in-page tab navigation to switch to an OS other than the OS you're currently running
3. Copy the URL
4. Open that URL in a new tab

**Expected:**
The newly opened install page honours the OS selection in the URL.
ie: Even on OS X / Windows, https://yarnpkg.com/en/docs/install#linux-tab will show the Linux tab.

**Actual:**
If https://yarnpkg.com/en/docs/install#linux-tab is visited on Windows, the page instead shows the Windows tab, and the URL in the location bar is reset to https://yarnpkg.com/en/docs/install#windows-tab

Note also that the location bar uses URLs of form `...#linux-tab` whereas the link target for the in-page navigation tabs uses `...#linux`. However regardless of which is used, the issue still occurs.

Many thanks :-)

Haroenv:
fixes #413

### DIFF
--- a/js/src/install.js
+++ b/js/src/install.js
@@ -1,4 +1,4 @@
 import { handleTabs, handleOS } from './lib/deeplink';
 
-handleTabs();
 handleOS();
+handleTabs();
--- a/js/src/lib/deeplink.js
+++ b/js/src/lib/deeplink.js
@@ -23,12 +23,14 @@ export function handleTabs() {
 }
 
 export function handleOS() {
-  // shows the tab of the current OS
-  [
-    [/Windows/i, '#windows-tab'],
-    [/Mac OS/i, '#mac-tab'],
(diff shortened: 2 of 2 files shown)


Haroenv:
First I checked if it was order of operations in  0232b2d67375bd5a40281420b9551ea1220db3ee, but actually the issue is that the redirecting to the current tab is always executed, even with the hash given. That's fixed in  374803a713ba1c3c14887ddc9264f8d2b91083d9

Daniel15:
Looks good, thanks!

edmorley:
Many thanks! :-)

PR #415, opened 2017-03-15
Updated install.md - added --ignore-engines option

Nexxado:
currently the --ignore-engines option is only documented as part of `yarn upgrade` command, but it is also valid for `yarn install`


Daniel15:
It's probably worth elaborating as to what this actually means. For example, what is an "engine", and what is being "checked"? For example, something like "Do not check that the current Node.js or Yarn version matches the range specified in the `engines` property in `package.json`" or something like that.

Nexxado:
True, but there is no elaboration under `yarn upgrade` docs as well.
I'm just matching what's already written there.

I made this pull request with the idea to save other people time, as I had an issue with `yarn install` failing due to the engines check and only after reading the `yarn upgrade` docs did I realize that the option to ignore engine checks even exists.

bestander:
@Nexxado, thanks for your contribution.
Could you rebase?

A paragraph as @Daniel15 suggested would help the project very much.
But even this note would be welcome

PR #417, opened 2017-03-18
feat(search) autoFocus SearchBox on search page

esphen:
# What
This PR adds a new feature - the `<SearchBox />` component now has the `autoFocus` property when the user loads the search page. 

# Why
When I arrive at the search page from an OpenSearch search in my browser, I would like to easily continue my search experience. Say I had a typo in my search, I would like to be able to correct it. Today I have to use my mouse to click the search field, then correct it and press enter. The mouse part seems redundant, so that is what this PR addresses.

See also https://github.com/yarnpkg/website/issues/405

### DIFF
--- a/js/src/lib/Search/index.js
+++ b/js/src/lib/Search/index.js
@@ -43,6 +43,7 @@ const Search = props => (
       ]}
     />
     <SearchBox
+      autoFocus={window.location.pathname.includes('/packages')}
       translations={{
         placeholder: window.i18n.search_placeholder,
       }}


js/src/lib/Search/index.js
```diff
@@ -43,6 +43,7 @@ const Search = props => (
       ]}
     />
     <SearchBox
+      autoFocus={window.location.pathname.includes('/packages/')}
```
Haroenv: should be `window.location.pathname.includes('/package'), because it doesn't include the trailing slash :smile:

you can see it live at https://deploy-preview-417--yarnpkg.netlify.com/en/packages

esphen: Wow, it's pretty cool that you have a test env for every PR. Love that.

Thanks for the review. I agree that we should drop the trailing slash and it should be `window.location.pathname.includes('/packages')`, but not `/package`. I don't think the correct behavior should be to highlight the search bar when you load a [search result][result] - at least that's not what I would expect. Having `/packages` would limit it to only the sarch screen.

Agree?

[result]: https://yarnpkg.com/en/package/babel

Haroenv: oh you're right, I mistyped, should be `/packages`

esphen: Amended the commit to remove the slash

PR #418, opened 2017-03-19
feat(documentation) add solano to ci docs

kyleholzinger:
# What
Add documentation around using yarn with SolanoCI to the CI page

# Why
So that people are aware of Solano and their native support of yarn


Daniel15:
Sounds good. I've never heard of Solano before.

kyleholzinger:
I think they're pretty small, but they're cool. Been using them for around a year and a half now and have been pretty satisfied

PR #421, opened 2017-03-21
fix(ci) solano

Haroenv:
1. file name wasn't *.md
  * caused jekyll/netlify to break and not deploy to production
2. Solano wasn't marked as string to translate
  * even if the deploy worked, it would still not have shown up before this


Daniel15:
Thanks! I wonder why this wasn't caught in the original PR. I expected Netlify to show a warning or something :/

PR #425, opened 2017-03-24
Add Scoop to list of Install options for Windows

tech4him1:
[Scoop](https://github.com/lukesampson/scoop) is a simple, developer-focused package manager for Windows.

### DIFF
--- a/lang/en/docs/_installations/windows.md
+++ b/lang/en/docs/_installations/windows.md
@@ -1,7 +1,7 @@
 
 ### Windows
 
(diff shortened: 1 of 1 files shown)


Daniel15:
Thanks, seems reasonable!

We probably need to automate updates of the Scoop package, similar to what we want to do for the Homebrew one (https://github.com/yarnpkg/yarn/issues/2841)

tech4him1:
Sounds good, let me know if you need any help with that.

PR #428, opened 2017-03-25
feat(config): put keys in a separate file

### LINKED ISSUE #426: pull out configuration variables into their own file
Algolia: 

* search only api key
* appid
* index name

Haroenv:
these keys aren't confidential, since you can only search with it.

fixes #426


Daniel15:
Thanks :D

PR #429, opened 2017-03-25
fix(i18n): provide a default value for scripted langs

### LINKED ISSUE #427: provide a default for the strings to translate in JS
Since operations are done on the strings, which are `undefined`, this throws errors for untranslated strings

Haroenv:
Because in some places operations like `.replace()` are used on the strings, and this doesn't work on `null`, languages that weren’t fully translated gave errors.

What’s done is to give them a default value of the English version of the text (which is a better experience anyway). 

Meanwhile I separated the values used in script from those not used in script.

(to be checked in the deploy preview, since i18n is very hard to test locally) [link](http://deploy-preview-429--yarnpkg.netlify.com/en/)

fixes #427

### DIFF
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -278,38 +278,41 @@ ci_semaphore: Semaphore
 ci_solano: Solano
 
 search_docs: Search documentation
-search_placeholder: Search packages (i.e. babel, webpack, react…)
-search_by_algolia: Search by Algolia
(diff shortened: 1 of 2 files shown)


Haroenv:
It works now, languages that don’t have the correct i18n yet render without errors (in English) 😄

Daniel15:
Thanks! This seems reasonable to me.

PR #430, opened 2017-03-26
Fixed typo

pborreli:


Daniel15:
Thanks! These instructions are very outdated though, so we'll definitely need to revise them at some point. The entire build process is automated now, including the Chocolatey package build.

PR #431, opened 2017-03-26
fix(redirects): explicitly route /packages/ to /packages/

Haroenv:
For some reason, Netlify interprets /en/packages/ to match /:lang/package/*, so this here will override that behaviour

(waiting on the deploy preview to see if it does the job)

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -2,6 +2,7 @@
 layout: null
 ---
 /:lang/package/* /lang/:lang/package 200
+/:lang/packages/ /lang/:lang/packages 200
 
 /downloads/:version/:file https://github.com/yarnpkg/yarn/releases/download/v:version/:file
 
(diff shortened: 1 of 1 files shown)


Haroenv:
Doesn’t work 😢

Haroenv:
closing for now, since it doesn’t seem to help

Haroenv:
This is working now; see https://fix-redirect-packages--algolia-yarn.netlify.com/en/packages

quote from Netlify support:

>Hi there,

>Sorry for the slow reply - had to check in with a guru on this one.  He found the bug and is working on a fix for it, but in the meantime, if you put that redirect at the very top of the file when you deploy, it should work even before we roll the bugfix out.

>Let me know in case that isn't the case!

vvo:
👌

Daniel15:
Thanks!

jdalton:
Awesome! I :heart: Netlify folks.

PR #432, opened 2017-03-26
fix(docsearch): custom styling

Haroenv:
1. layout for the searchbox to take up all the space on layouts smaller than `-lg`
2. apply custom docsearch style again
  - this was enabled in the December version, but overriden by docsearch. Now the selector issues have been fixed and the colours have been made in such a way that they apply correctly. 

before|after
---|---
screen shot 2017-03-26 at 15 02 16 | screen shot 2017-03-26 at 15 01 57


Daniel15:
Thank you! 🐈

PR #434, opened 2017-03-26
feat(search): mark count location in download string

Haroenv:
Just like the other strings, the number can be in different locations based on teh language. The number will be displayed using Number.toLocaleString(), so it's localised itself


Daniel15:
Thank you!

PR #435, opened 2017-03-27
fix broken liquid comment

gnujoow:
there is broken comment showing up in index page.

it can be written in js comment `//` or liquid comment `{% comment %} {% endcomment %}`. but I thought it was originally written in liquid so I fixed it like it :)


Daniel15:
Good catch! cc @Haroenv - I'm not quite sure how this happened.

Haroenv:
Ah that's because I was doing fancy editor regex tricks to make that transform work. Good catch!

PR #437, opened 2017-03-27
Adding example of cli flag for dependency categories

cdesch:
Adding reference examples of adding a dependency to different categories of dependencies as outlined in the MAN pages of Yarn


Daniel15:
This wording assumes that the person already knows what "devDependencies" and "peerDependencies" are. I'd suggest having separate headings and explaining them individually.

bestander:
Let's merge it as is but a clarification as @Daniel15  suggested will be very useful.

PR #440, opened 2017-03-28
fix(detail): make changing languages work

### LINKED ISSUE #439: Changing language while on /package/xxx links to /package/
This is because the language dropdown is rendered by Jekyll, which doesn't have knowledge about the frontend route. 

This can be fixed by doing a `.forEach` over the language dropdown (like in production.js) and replace the link by /package/xxx.

Haroenv:
Since the url isn't aware of which package detail we're on, we need to set the url to the correct package.

Doesn't work in `dev`, because `dev` doesn't have i18n

fixes #439 

PR deploy: https://deploy-preview-440--yarnpkg.netlify.com/zh-Hans/package/test

### DIFF
--- a/js/src/package.js
+++ b/js/src/package.js
@@ -11,7 +11,22 @@ const id = process.env.NODE_ENV == 'production'
   ? location.pathname.split('/').slice(3).join('/')
   : location.search.substring(1);
 
+if (!NodeList.prototype.forEach) {
+  NodeList.prototype.forEach = Array.prototype.forEach;
+}
+
+function languageDropdownAddPackage(pkg) {
+  const langMenu = document.getElementById('dropdownNavLanguageMenu');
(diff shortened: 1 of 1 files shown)


js/src/package.js
```diff
@@ -11,7 +11,22 @@ const id = process.env.NODE_ENV == 'production'
   ? location.pathname.split('/').slice(3).join('/')
   : location.search.substring(1);
 
+if (!NodeList.prototype.forEach) {
+  NodeList.prototype.forEach = Array.prototype.forEach;
```
Daniel15: It seems unnecessary to extend the prototype... Just do a regular `for` loop instead.

Haroenv: Sure. We're extending the prototype for the same already in production.js though

Haroenv: Changed it to `for` now 😄

PR #443, opened 2017-04-02
feat(details): add devDependencies and package.json

Haroenv:
This is useful metadata and make the 'Usage' section not as small.

TODO:

- [x] add to Algolia index
- [x] new icons (cc @Shipow)
- [x] wording of package.json

before|after
---|---
screen shot 2017-04-02 at 12 53 15 |


js/src/lib/Details/Aside.js
```diff
@@ -6,12 +6,31 @@ import Activity from './Activity';
 import Popularity from './Popularity';
 import Usage from './Usage';
 import { Owner } from '../Hit';
+import { encode, prefixURL } from '../util';
+
+const packageJSONLink = ({ githubRepo, gitHead }) => {
```
Daniel15: Just use regular function syntax here
```
function packageJSONLink({ githubRepo, gitHead }) {
```

The arrow syntax is longer and doesn't really provide any added value.

Haroenv:
Done now!

http://deploy-preview-443--yarnpkg.netlify.com/en/package/jest for example

screen shot 2017-04-03 at 14 34 01

js/src/lib/Details/index.js
```diff
@@ -54,18 +54,12 @@ class Details extends React.Component {
...
-              base: 'https://raw.githubusercontent.com',
-              user: this.state.githubRepo.user,
-              project: this.state.githubRepo.project,
-              head: this.state.gitHead ? this.state.gitHead : 'master',
-              path: this.state.githubRepo.path.replace(/\/tree\//, ''),
-            }),
-        type: 'text',
-      }).then(res => this.setState({ changelog: res }));
+      if (this.state.changelogFilename) {
+        get({
```
Haroenv: this is actually kind of separate, but it could've been changed earlier. Basically at indexing time (our backend), we check a lot of possible file names, and CHANGELOG.md is always tested as well, so a fallback isn't needed in this case.

lang/en/package.html
```diff
@@ -17,66 +17,61 @@ <h2 class="details-main--title d-inline-block m-2">…</h2>
...
-        <div class="copyable"><code class="copyable--code"><!-- react-text: 31 -->$ <!-- /react-text --><span>yarn add …</span></code><button class="copyable--button"><img src="/assets/detail/ico-copy-default.svg" alt="" class="copyable--button__img"><!-- react-text: 35 -->copy<!-- /react-text --></button></div>
+        <div class="copyable"><code class="copyable--code"><!-- react-text: 30 -->$ <!-- /react-text --><span>yarn add …</span></code><button class="copyable--button"><img src="/assets/detail/ico-copy-default.svg" alt="" class="copyable--button__img"><!-- react-text: 34 -->copy<!-- /react-text --></button></div>
         <div><a class="details-side--runkit" href="https://runkit.com/npm/…" target="_blank" rel="noopener noreferrer">Try in RunKit</a></div>
       </article>
       <article class="details-side--popularity">
         <h1>Popularity</h1>
         <dl></dl>
       </article>
-      <article class="details-side--activity">
-        <h1>Activity</h1>
```
Haroenv: This doesn't show anymore in the detail page when there's no GitHub repo, so it doesn't show by default

js/src/lib/Details/Usage.js
```diff
@@ -1,27 +1,49 @@
...
-const Usage = ({ dependencies }) => (
+const Deps = ({ dependencies, text, id }) =>
+  dependencies
+    ? <div className="d-flex flex-items-between w-100">
+        <img src={`/assets/detail/ico-${id}.svg`} alt="" />
+        <dt>
+          {Object.keys(dependencies).length > 0
+            ? <details>
+                <summary>{text}</summary>
+                {Object.keys(dependencies)
```
Daniel15: This is calling `Object.keys` there times, which will iterate over the object unnecessarily. Can you just call it once and cache it in a variable (like `const deps = Object.keys(dependencies)`?

js/src/lib/Details/Usage.js
```diff
@@ -1,27 +1,49 @@
...
-                  <summary>{window.i18n.detail.dependencies}</summary>
-                  {Object.keys(dependencies)
-                    .map((name, index) => (
-                      <a href={packageLink(name)} key={index}>{name}</a>
-                    ))
-                    .reduce((prev, curr) => [prev, ', ', curr])}
-                </details>
-              : 'Dependencies'}
-          </dt>
+          <img src={`/assets/detail/ico-package-json.svg`} alt="" />
```
Daniel15: Just do `src="/assets/detail/ico-package-json.svg"`

lang/en/package.html
```diff
@@ -17,66 +17,61 @@ <h2 class="details-main--title d-inline-block m-2">…</h2>
             <!-- react-text: 8 -->
             <!-- /react-text -->
           </a><span class="ais-Hit--popular null" title="0 downloads in the last 30 days">0</span>
-          <!-- react-empty: 10 --><span class="ais-Hit--version">0.0.0</span></div>
+          <!-- react-empty: 10 --><span class="ais-Hit--version"></span></div>
```
Daniel15: Is this file generated through a script? I'm guessing it's the result of rendering the React component to a string (given the presence of ` <!-- react-empty: 10 -->` and such) but it'd be good to have an easy way to keep it up-to-date.

Haroenv: Not really through a script, I just load the view without any package manually and copy the html that's there. 

Do you think it's worth it to write a script?

Daniel15:
Looks good! Thanks!

lang/en/package.html
Daniel15: I think it'd be nice to have a script. Then we could include it as part of the build, ensuring that it's always up-to-date (I'm not a fan of committing build artifacts to source control).

PR #444, opened 2017-04-04
docs(install): list the version on the installation page

### LINKED ISSUE #252: Prominently display latest version number somewhere
As mentioned in https://github.com/yarnpkg/yarn/issues/1730

It's currently non-trivial to figure out what the latest version of yarn is from the website, it'd be good to throw this number in the header somewhere.

It would also be good to provide some guidance during or near the installation instructions about how to upgrade `yarn` to the latest version - does `self-update` work regardless of the install method?

Haroenv:
closes #252

It's now possible to add any `hero_subtext` include by listing the _includes filename in the frontmatter of a layout or a page.

see http://deploy-preview-444--yarnpkg.netlify.com/en/docs/install/

### DIFF
--- a/_includes/version.html
+++ b/_includes/version.html
@@ -1,0 +1,17 @@
+<div class="mt-4">
+  {{i18n.site_nav_stable_version}}:
+  <strong class="navbar-text">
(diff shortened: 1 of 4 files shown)


Daniel15:
Nice!

PR #454, opened 2017-04-16
docs(install): update Arch Linux section

michael-zapata:
Update from the tool-dependent command to the standard installation step

The package is available in the official repositories since this commit:
https://git.archlinux.org/svntogit/community.git/commit/?id=1b135175a47e20d9c9fb7ccd0baec205178a9889


Daniel15:
Thanks!

PR #455, opened 2017-04-19
Fix relative/broken links in docs

ndresx:
Stumbled upon the following two pages in the docs which contain a couple of relative/broken links when the pages are requested with a trailing slash. I have fixed these links to become absolute ones in order to be consistent with the rest of the pages.

- https://yarnpkg.com/en/docs/cli/unlink/
- https://yarnpkg.com/en/docs/installing-dependencies/

This problem is not noticeable when browsing normally on the Yarn website, because the whole navigation is working (and linked) without trailing slashes, although the `canonical` tags and therefore indexed pages always contain the trailing slash.

The localized canonicals are also a bit inconsistent. While the English docs landing page for example is reachable via `/en/docs/` _(with or without trailing slash)_, its canonical is pointing to `/lang/en/docs/`.

Maybe it might be better to reduce the number of possibilities of how to access these pages, but that's another issue; just wanted to mention this.

### DIFF
--- a/lang/en/docs/cli/unlink.md
+++ b/lang/en/docs/cli/unlink.md
@@ -4,9 +4,11 @@ guide: docs_cli
 layout: guide
 ---
 
+{% include vars.html %}
+
 <p class="lead">Unlink a previously created symlink for a package.</p>
 
-To remove a symlinked package created with [`yarn link`](./link), `yarn unlink` can be used.
+To remove a symlinked package created with [`yarn link`]({{url_base}}/docs/cli/link), `yarn unlink` can be used.
 
 ##### `yarn unlink` <a class="toc" id="toc-yarn-unlink" href="#toc-yarn-unlink"></a>
 
@@ -18,7 +20,7 @@ To unlink a package that was symlinked during development in your project, simpl
 run `yarn unlink [package]`. You will need to run `yarn` or `yarn install` to re-install
 the package that was linked.
 
-Continued example from the [`yarn link`](./link) documentation: assume two folders
+Continued example from the [`yarn link`]({{url_base}}/docs/cli/link) documentation: assume two folders
 `react` and `react-relay` that are located next to each other with `react` linked
 into the `react-relay` project:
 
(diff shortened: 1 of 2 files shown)


lang/en/docs/cli/unlink.md
```diff
@@ -4,9 +4,11 @@ guide: docs_cli
 layout: guide
 ---
 
+{% include vars.html %}
```
Haroenv: I’m pretty sure that this is already included in `default.html`, but should be checked

lang/en/docs/cli/unlink.md
```diff
@@ -4,9 +4,11 @@ guide: docs_cli
...
+
 <p class="lead">Unlink a previously created symlink for a package.</p>
 
-To remove a symlinked package created with [`yarn link`](./link), `yarn unlink` can be used.
+To remove a symlinked package created with [`yarn link`]({{url_base}}/docs/cli/link), `yarn unlink` can be used.
```
Haroenv: Yep, that’s more consistent with the others (while the previous version also worked)

lang/en/docs/installing-dependencies.md
```diff
@@ -5,13 +5,15 @@ layout: guide
 additional_reading_tags: ["dependencies", "package-json", "yarn-lock", "cli-install"]
 ---
 
-If you have just checked out a package from [version control](./version-control), you will need to install those dependencies.
+{% include vars.html %}
```
Haroenv: same, this is probably not necessary, or else it should be added to `_layouts/guide`

lang/en/docs/cli/unlink.md
ndresx: You are right, it is. I basically just copied over the same requirements from the other md files in the cli folder (which all contain this `include`). I can remove it if you want.

lang/en/docs/cli/unlink.md
ndresx: For me it doesn't work if you open https://yarnpkg.com/en/docs/cli/unlink/. Because this link is a relative one as described in the description, it will just get appended and then look like this https://yarnpkg.com/en/docs/cli/unlink/link => 404.

lang/en/docs/installing-dependencies.md
ndresx: Yup, not necessary, just the same format as the other md files in this folder. As said in my previous comment, I can remove it if you want.

PR #457, opened 2017-04-21
Update global.md

mscolnick:


Daniel15:
I don't think this change is right... The text is explaining that `yarn add global ...` doesn't globally add it 😛

mscolnick:
Oh ya, oops. I misread :/

PR #463, opened 2017-04-23
Update unix_path_setup.md

zky829:
``export PATH="$PATH:`yarn global bin`"` doesn't work or make sense as a step right after installing/extracting yarn. Feel free to correct/update the wording of the update!


zky829:
Duplicate PR.

PR #465, opened 2017-04-25
changed 'yarn install --force' to 'yarn add --force' in docs/migratin…

digitalcrafted:
updated documentation in migrating from npm to use 'yarn add --force' instead of  'yarn install --force' for 'npm rebuild'


bestander:
I don't think it is correct.
`yarn install --force` will reinstall dependencies, `yarn add` requires an argument which dependency to add

PR #467, opened 2017-04-25
Add yarn install offline flag

Sunyang730:
- Add offline as a flag for yarn install docs


Sunyang730:
@bestander Also update the offline flag on the docs

PR #478, opened 2017-05-03
Improve the Search results

Haroenv:
1. Use infinite hits instead of pagination
  - the actual page number doesn't really matter much
  - move the page number out of the URL
    - UX for landing on a page is weird
    - we didn't use that parameter in showing the results
2. Show number of results
  - more useful than the pagination, since it's a number
  - also shows the ms it takes to do this request 

https://deploy-preview-478--yarnpkg.netlify.com/en/

### DIFF
--- a/_sass/_search.scss
+++ b/_sass/_search.scss
@@ -108,32 +108,6 @@ $ais-muted-color: rgba(black,.5);
   }
 }
 
-.ais-Pagination {
-  &__root {
-    margin: 2em auto;
-    @extend .pagination;
-  }
-
-  &__item {
-    @extend .page-item;
-  }
-
-  &__itemLink {
-    @extend .page-link;
-  }
-
-  &__itemSelected {
-    @extend .active;
-  }
-
-  &__itemDisabled {
-    @extend .disabled;
-    visibility: visible;
-    opacity: .3;
-    cursor: default;
-  }
-}
-
 .ais-Hits {
   &--item {
     padding: 1.5rem 1rem 2rem;
@@ -148,6 +122,19 @@ $ais-muted-color: rgba(black,.5);
   &__empty {
     padding: 2.5rem 0 2rem;
     border-bottom: 1px solid $gray-lighter;
+  }
+}
+
+.ais-InfiniteHits {
+  &__root {
+    display: flex;
+    flex-direction: column;
+  }
+
+  &__loadMore {
+    margin: 1em auto;
+    @extend .btn;
+    @extend .btn-primary;
   }
 }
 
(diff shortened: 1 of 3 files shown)


bestander:
A use case for pagination is when you go to package info and return back, it is very sad to lose your position and do "load more" many times every time.

Haroenv:
That's right, but don't you want a `load previous` button in that case too?

bestander:
I think that it should load automatically.
My point is that pagination is way more back button friendly than infinite scroll.
I can't speak for all users but I switch between search and package info a lot when looking for a right one

Haroenv:
You're right, this is actually a React InstantSearch issue, I'll take a look how it's possible.

Haroenv:
So unfortunately this won't be easy, it would require major changes in how InfiniteHits is written. However I can enable the `&p=` syntax again . This will load the nth page when that link is shared. I'm not sure if that is the best solution, because sharing a certain page of the results is usually not what you want, rather just the query. 

However, I noticed something interesting when putting the `&p=` back in the url, that there was a bug in the current implementation. compare https://yarnpkg.com/en/packages?q=test&p=6 with https://yarnpkg.com/en/packages?q=test&p=2 or https://yarnpkg.com/en/packages?q=test&p=0 for example. This might mean that the problem you're legitimately raising might not actually come up in real usage, since nobody said something about it.

Regarding your comment on being able to press the back button and still being at the same place, in my testing on the [deploy preview](https://deploy-preview-478--yarnpkg.netlify.com/en/packages?q=test), clicking on a result and then the back button will still have the DOM and scroll position in view, so unless your browser crashes, that situation is already handled. 

Almost every page that has infinite scroll doesn't share pagination though (Facebook, Twitter, Pinterest), so I think this is the way to go for now.

bestander:
If back button scenario is supported in the infinite scroll mode and a user can go back and forward between info page and a search result (after hitting Load More a few ties) then it sounds good to me.

Does it perform well?
For example, searching for "bluetooth" returns 63 pages, I literally had to go through a big chunk of those once.

Social networks engage with users in a different way.
When you view a news feed you see detailed results without a need to leave the feed.
And also the prioritization in a news feed is not persistent, so sharing a page 2 of your feed won't be the same at different time or for a different user.

It would be correct to compare yarn website search to google search where you have pagination.

Overall, I don't object infinite scroll, go with it if you feel strong that it improves UX.

Haroenv:
In the end after thinking a lot about it and trying lots of strategies, I have the feeling that infinite pagination is not ready to use in this context yet. I went back to the regular pagination, so all this PR does is add the "stats", so I'll merge this.

bestander:
👍

PR #481, opened 2017-05-07
Use treeshook version of react-instantsearch

Haroenv:
React InstantSearch now publishes es2015 modules, and thus the advantage for writing the content of the respective files isn't really there anymore. This makes the code simpler, as well as a few kb's smaller 🚀


Daniel15:
Thanks!

PR #485, opened 2017-05-09
Correct warning about when npm is non-deterministic

forivall:
Also made a small grammatical improvement.

Revision of #484


Daniel15:
Thank you! 😃

PR #494, opened 2017-05-14
feat(files): show size in file listing

Haroenv:
For this to make sense I added a background on hovering a certain listing, since they might sometimes be pretty wide.

I also added another utility class that should be in bootstrap, but isn't for some reason, but it's to align on the baseline.

I used the small [bytes](https://yarnpkg.com/en/package/bytes) utility to show the sizes in kB etc.

Screenshot:

screen shot 2017-05-14 at 17 30 50

[preview](https://deploy-preview-494--yarnpkg.netlify.com/en/package/algoliasearch-helper)

### DIFF
--- a/_sass/_details.scss
+++ b/_sass/_details.scss
@@ -132,6 +132,20 @@
       & & {
         padding-left: 1.5em;
       }
+
+      li {
+        padding: .2em;
+
+        &:hover {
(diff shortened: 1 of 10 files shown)


Daniel15:
Nice!

PR #497, opened 2017-05-15
Document yarn install --check-files

raido:
Add documentation for `yarn install --check-files`.

### DIFF
--- a/lang/en/docs/cli/install.md
+++ b/lang/en/docs/cli/install.md
@@ -23,6 +23,10 @@ Running `yarn` with no command will run `yarn install`, passing through any prov
 Install all the dependencies listed within `package.json` in the local
 `node_modules` folder.
 
+##### `yarn install --check-files` <a class="toc" id="toc-yarn-install-check-files" href="#toc-yarn-install-check-files"></a>
+
+Verifies that already installed dependencies match with yarn's lock file.
+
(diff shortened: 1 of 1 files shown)


lang/en/docs/cli/install.md
```diff
@@ -23,6 +23,10 @@ Running `yarn` with no command will run `yarn install`, passing through any prov
 Install all the dependencies listed within `package.json` in the local
 `node_modules` folder.
 
+##### `yarn install --check-files` <a class="toc" id="toc-yarn-install-check-files" href="#toc-yarn-install-check-files"></a>
+
+Verifies that already installed dependencies match with yarn's lock file.
```
bestander: Better 
```
Verifies that already installed files in node_modules did not get removed
```

This feature does not check with the lockfile but rather reads the files array from node_modules/.yarn-integrity file and checks that all files are in place

bestander:
Thanks!

lang/en/docs/cli/install.md
XOder64x: > Better
> 
> ```
> Verifies that already installed files in node_modules did not get removed
> ```
> 
> This feature does not check with the lockfile but rather reads the files array from node_modules/.yarn-integrity file and checks that all files are in place

PR #504, opened 2017-05-18
Add post announcing new custom-host-suffix option

lumaxis:
As discussed with @bestander in https://github.com/yarnpkg/yarn/pull/3231, I added a quick post explaining the new feature.

I don't really know what the guidelines are for the posts on your blog so I tried to align myself to some of the previous ones. Feel free to criticize or change anything as it makes sense to you 👍

### DIFF
--- a/_posts/2017-05-19-supporting-more-registries.md
+++ b/_posts/2017-05-19-supporting-more-registries.md
@@ -34,7 +34,7 @@ As you can see, the archive's URL does not contain the actual feed name anymore 
 
 ### Introducing custom host suffixes
 
-Starting with version `0.0.26`, Yarn now understands a new configuration option called `custom-host-suffix`. This allows you to keep the same strict URL validations for most of your package URLs but also selectively loosen that check for a specific registry provider so that Yarn will now match the URLs where the host part ends with the value from this new option.
+Starting with version `0.26.0`, Yarn now understands a new configuration option called `custom-host-suffix`. This allows you to keep the same strict URL validations for most of your package URLs but also selectively loosen that check for a specific registry provider so that Yarn will now match the URLs where the host part ends with the value from this new option.
 
 Simply add `custom-host-suffix` to either your global user-level `.npmrc` or your project's individual `.npmrc` and Yarn will be able to download your packages as desired.
 In the above example of Visual Studio Team Services, the `.npmrc` should contain an entry like this:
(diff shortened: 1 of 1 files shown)


_posts/2017-05-19-supporting-more-registries.md
```diff
@@ -0,0 +1,46 @@
...
+https://$ACCOUNT_NAME.pkgs.visualstudio.com/_packaging/da6e033f-20ad-4ee1-a784-8995dd6836b72/npm/registry/@scope/package-name/-/package-name-0.0.1.tgz
+```
+
+As you can see, the archive's URL does not contain the actual feed name anymore but rather a random GUID, followed by the package name and version. This differing layout of the path part of the URL lead to Yarn not recognizing that the two URLs actually both do belong to a request to the same registry and would therefore refuse to download the package.
+
+### Introducing custom host suffixes
+
+Starting with version `0.26.0`, Yarn now understands a new configuration option called `custom-host-suffix`. This allows you to keep the same strict URL validations for most of your package URLs but also selectively loosen that check for a specific registry provider so that Yarn will now match the URLs where the host part ends with the value from this new option.
+
+Simply add `custom-host-suffix` to either your global user-level `.npmrc` or your project's individual `.npmrc` and Yarn will be able to download your packages as desired.
```
bestander: How about `.yarnrc` with an example?
```
yarn config set custom-host-suffix <url>
```

lumaxis: @bestander The `custom-host-suffix` actually does not work with `.yarnrc` right now, as far as I can tell.
I think `npm-registry.js` only loads config values from `.npmrc` right now: https://github.com/yarnpkg/yarn/blob/master/src/registries/npm-registry.js#L132

bestander: Yeah but there us yarn-registry that inherits from npm-registry and reads .yarnrc.
Could you please check if yarnrc works as well?
I expect that it should work.

atti187: Can confirm - doesn't work to add custom-host-suffix to .yarnrc with 0.26.0-20170524.1610. Had to put it into .npmrc and haven't investigated further.

bestander: Ok, let's merge as is and iterate further

bestander:
Thanks, @lumaxis!

gsklee:
@bestander Maybe modify the post date so it can come on top as the newest post?

bestander:
Go ahead, send a PR

lumaxis:
@gsklee @bestander Added #530

PR #505, opened 2017-05-19
Update install and remove commands doc

raido:
Add --ignore-scripts to install command docs.

Add note about flags passed from remove command to install command. <-- Not sure if this should be promoted, if not then I am happy to remove it.


bestander:
Thanks, @raido, docs updates are always welcome.
I did a few clarifications though, let me know if you meant a different thing.

PR #512, opened 2017-05-24
Update 2016-11-24-offline-mirror.md

Sunyang730:
This is what I had encountered and this is a workaround for updating node modules


Sunyang730:
@bestander take a look

bestander:
Thanks, @Sunyang730

PR #513, opened 2017-05-26
Upgrade docs to add `yarn upgrade --latest` flag

rally25rs:
This is the documentation change that corresponds to https://github.com/yarnpkg/yarn/pull/3510
They should be merged at the same time.


bestander:
Thanks, @rally25rs!

PR #516, opened 2017-05-29
Add homebrew install instructions for nvm compat

ntwb:
Fixes https://github.com/yarnpkg/yarn/issues/3255

### DIFF
--- a/lang/en/docs/_installations/mac.md
+++ b/lang/en/docs/_installations/mac.md
@@ -9,6 +9,12 @@ This will also install Node.js if it is not already installed.
 brew install yarn
 ```
 
+If you use [nvm](https://github.com/creationix/nvm) you should exclude installing NodeJS so that the nvm binaries are used.
+
+```sh
+brew install yarn --ignore-dependencies
+```
+
 #### MacPorts
 
 You can install Yarn through [MacPorts](https://www.macports.org/).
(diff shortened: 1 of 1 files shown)


lang/en/docs/_installations/mac.md
```diff
@@ -9,6 +9,12 @@ This will also install Node.js if it is not already installed.
 brew install yarn
 ```
 
+If you use [nvm](https://github.com/creationix/nvm) you should exclude installing NodeJS so that the nvm binaries are used.
```
Daniel15: Maybe change "nvm binaries" to "nvm's version of Node.js"?

Also, small nit: Use "Node.js", not "NodeJS".

ntwb: Ha, and done, I've been writing it that way for years, my bad, sorry Node.js

Daniel15:
Thanks!

nandorojo:
Is this no longer published on the docs? Looks like the current installation explanation is a little less clear https://classic.yarnpkg.com/en/docs/install/#mac-stable

Thanks!

Daniel15:
@nandorojo  Homebrew removed the ability to install Yarn without installing Node.js a while back. In fact they removed all customization options for formulae.

nandorojo:
You’re right. I went with the CURL install instead — this may be the best option for nvm users to avoid node installs. Maybe I could add a PR to suggest that?

Daniel15:
Sure, feel free to submit a PR.

PR #517, opened 2017-05-30
Adds install instructions for alpine linux

errm:


bestander:
Thanks, @errm

errm:
wow that was super quick, thanks :)

PR #518, opened 2017-05-31
Reword npm installation warning to avoid wrong implications

gaearon:
A version of this was originally submitted in https://github.com/yarnpkg/website/pull/484 and closed in favor of https://github.com/yarnpkg/website/issues/485 which got merged.

I think we can further improve it to make it clear this is not a dig at npm, and that the problem is common to any mainstream Node-based package managers. I kept the title as `Install via npm` and kept the mention of npm in the first sentence because that's realistically what people are going to do, but I hope this makes it clearer that the issue is not in npm itself, but rather that today’s Node package managers don’t give these guarantees (Yarn being no exception).

### DIFF
--- a/lang/en/docs/_installations/alternatives.md
+++ b/lang/en/docs/_installations/alternatives.md
@@ -12,7 +12,7 @@ recommended to install Yarn via our packages instead.
 #### Install via npm
 
 > **Note:** Installation of Yarn via npm is generally not recommended.
-> Installing Yarn with npm is non-deterministic, the package is not signed,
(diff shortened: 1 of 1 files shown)


Daniel15:
Thanks, @gaearon. This looks like a reasonable change to me. In theory, we could have a "Node-based package manager" that performs proper integrity checks, but I guess we could just update the text again if that were to ever happen.

gaearon:
If one appears, we can say "We don't suggest installing Yarn with Node-based package manager except X because ...", or even suggest to use it.

PR #520, opened 2017-06-01
feat(deprecation): show deprecated message

Haroenv:
The `deprecated` value in the index is either `false` or a string. This adds that string as a `title` to that badge. In the future this message would be visible more clearly, but I couldn't think of a good UI to put it, so this is a nice inbetween state where you can already read the message if you want

related: https://github.com/algolia/yarnpkg-website/issues/20#issuecomment-304726188

cc @vvo

### DIFF
--- a/js/src/lib/Hit/index.js
+++ b/js/src/lib/Hit/index.js
@@ -15,7 +15,9 @@ export const License = ({ type }) =>
 
 export const Deprecated = ({ deprecated }) =>
   deprecated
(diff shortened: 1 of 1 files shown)


vvo:
I would put the message instead of the package description in search results. (You should now very soon that's its deprecated).
I would put the message in bold on top of the description on the package page.

WDYT?

Haroenv:
Seems like a good addition, will do that later 😄

PR #522, opened 2017-06-01
feat(i18n): add translation key for stats

Haroenv:
before|after
---|---
screen shot 2017-06-01 at 16 02 59|screen shot 2017-06-01 at 15 52 25

the number will use the locale of the current url, as in other places


Haroenv:
Slightly different copy, “packages” vs the generic “results”, as well as enabling it to be translated on crowdin

PR #528, opened 2017-06-15
Check apt-transport-https availability

4oXdev:


Daniel15:
Rather than adding it to the block of commands, I'd suggest adding it below
as a side note. Something like "if you encounter an error while running
'apt-get update', you may need to install apt-transport-https".

4oXdev:
@Daniel15 look it https://github.com/yarnpkg/website/pull/528/commits/36570e09d321ccd851b321c7c793f858544881ff

PR #530, opened 2017-06-16
Update date in Private Registry Support post

lumaxis:
This also renames the file, I hope this doesn't break anything as this is not yet published?


bestander:
Thanks, @lumaxis!

PR #533, opened 2017-06-18
fix(markdown): rewrite urls to GitHub relative ones in raw html

### LINKED ISSUE #531: images in readme don't get displayed for package prettier
See https://github.com/prettier/prettier/issues/2186

https://yarnpkg.com/en/package/prettier

27258194-98a9efb8-5437-11e7-9a5a-844ef6dbf401

cc @Haroenv

Haroenv:
Raw html doesn't get lexed and parsed by Marked, and thus the custom rules for images and links don't get applied. To fix that, the same rewriting needs to happen on raw html too.

I might have a little duplicate code here, I wonder how to make it simpler, but I'm pretty tired now and this works at least.

closes prettier/prettier#2186
closes #531

note: might want to switch to something that works nicer with React than marked, and has the same rewriting and GFM capabilities, but not worth the effort yet

### DIFF
--- a/js/src/lib/Details/Markdown.js
+++ b/js/src/lib/Details/Markdown.js
@@ -4,6 +4,11 @@ import xss from 'xss';
 
 import { prefixURL } from '../util';
 
+const GITHUB = {
+  main: 'https://github.com',
+  raw: 'https://raw.githubusercontent.com',
+};
+
 marked.Lexer.rules.gfm.heading = marked.Lexer.rules.normal.heading;
 marked.Lexer.rules.tables.heading = marked.Lexer.rules.normal.heading;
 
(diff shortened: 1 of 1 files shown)


Haroenv:
http://deploy-preview-533--yarnpkg.netlify.com/en/package/prettier

vjeux:
Thanks for the quick fix!

js/src/lib/Details/Markdown.js
```diff
@@ -12,24 +17,31 @@ const renderAndEscapeMarkdown = ({ source, githubRepo, gitHead }) => {
...
-        path,
-      })}" title="${title}">${text}</a>`;
+    renderer.image = (href, title, text) =>
+      `<img src="${prefix(href, GITHUB.raw)}" title="${title}" alt="${text}"/>`;
+
+    renderer.link = (href, title, text) =>
+      `<a href="${prefix(href, GITHUB.main)}" title="${title}">${text}</a>`;
+
+    renderer.html = function(html) {
+      return html
```
Daniel15: It'd be a lot cleaner to use a Markdown parser that returns an AST, so that we could modify this directly in the AST rather than using string replacement. Something like [Markdig](https://github.com/lunet-io/markdig) but for JavaScript. Does an advanced Markdown parser like that exist in JavaScript?

js/src/lib/Details/Markdown.js
```diff
@@ -12,24 +17,31 @@ const renderAndEscapeMarkdown = ({ source, githubRepo, gitHead }) => {
...
+        .replace(
+          /src="([^"]*)/g,
+          (match, href) => `src="${prefix(href, GITHUB.raw)}`
+        )
+        .replace(
+          /href="([^"]*)/g,
```
Daniel15: Doing two replacements with the exact same regex seems wasteful - Could you do both of these in a single `.replace` call, depending on the URL?

js/src/lib/Details/Markdown.js
BYK: Did you mean this :https://remarkjs.com/#1

:)

js/src/lib/Details/Markdown.js
Haroenv: Yeah, I wasn't sure if doing replace twice or an if was better

js/src/lib/Details/Markdown.js
Haroenv: Ah, that one was out of the table earlier because it didn’t have GFM tables, but since 21 hours ago it does; @BYK

Haroenv: https://github.com/mapbox/remark-react seems useful here

PR #537, opened 2017-06-21
Update upgrade-interactive to use correct semver

kaylieEB:
Incorrect help option was being used [here](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/upgrade-interactive.js#L17). The default mode uses minor versions, and tilde should refer to patch (separate PR for help option here: https://github.com/yarnpkg/yarn/pull/3690)


bestander:
Thanks, @kaylieEB

PR #543, opened 2017-06-25
use sh shell instead of Bash

badersur:
When installing a [nightly build](https://yarnpkg.com/en/docs/nightly) using these commands:

```
wget https://yarnpkg.com/install.sh
chmod +x install.sh
./install.sh --nightly
```

yarn installs but I get this *annoying* messages:

```terminal
./install.sh: 48: ./install.sh: [[: not found
./install.sh: 53: [: unexpected operator
```

I know a little about shell scripting & this seems to be a [bash related thing](https://stackoverflow.com/questions/669452/is-preferable-over-in-bash) (and you use bash in your [installation script](https://yarnpkg.com/en/docs/install#alternatives-tab): `curl -o- -L https://yarnpkg.com/install.sh | bash`)

I'm suggesting changing sh shebang to bash shebang based on [This Stack Overflow question](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang).

I'm using Ubuntu 17.04.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ white="\033[37m"
 gpg_key=9D41F3C3
 
 yarn_get_tarball() {
-  printf "$cyan> Downloading tarball...$reset\n"
+  echo "${cyan}> Downloading tarball...${reset}"
   if [ "$1" = '--nightly' ]; then
     url=https://nightly.yarnpkg.com/latest.tar.gz
   elif [ "$1" = '--rc' ]; then
@@ -18,26 +18,26 @@ yarn_get_tarball() {
   elif [ "$1" = '--version' ]; then
     # Validate that the version matches MAJOR.MINOR.PATCH to avoid garbage-in/garbage-out behavior
     version=$2
-    if echo $version | grep -qE "^[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$"; then
+    if echo "$version" | grep -qE "^[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$"; then
       url="https://yarnpkg.com/downloads/$version/yarn-v$version.tar.gz"
     else
-      printf "$red> Version number must match MAJOR.MINOR.PATCH.$reset\n"
+      echo "${red}> Version number must match MAJOR.MINOR.PATCH.${reset}"
       exit 1;
     fi
   else
     url=https://yarnpkg.com/latest.tar.gz
   fi
   # Get both the tarball and its GPG signature
-  tarball_tmp=`mktemp -t yarn.tar.gz.XXXXXXXXXX`
+  tarball_tmp=$(mktemp -t yarn.tar.gz.XXXXXXXXXX)
   if curl --fail -L -o "$tarball_tmp#1" "$url{,.asc}"; then
-    yarn_verify_integrity $tarball_tmp
+    yarn_verify_integrity "$tarball_tmp"
 
-    printf "$cyan> Extracting to ~/.yarn...$reset\n"
+    echo "${cyan}> Extracting to ~/.yarn...${reset}"
     mkdir .yarn
-    tar zxf $tarball_tmp -C .yarn --strip 1 # extract tarball
-    rm $tarball_tmp*
+    tar zxf "$tarball_tmp" -C .yarn --strip 1 # extract tarball
+    rm $tarball_tmp* # Ignore shellcheck warning; it's intended to have globbing!
   else
-    printf "$red> Failed to download $url.$reset\n"
+    echo "${red}> Failed to download ${url}.${reset}"
     exit 1;
   fi
 }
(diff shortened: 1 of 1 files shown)


Daniel15:
We can't change it to Bash because then the script won't work on operating
systems that don't include Bash by default.

vvo:
We had to do the same and it's also what does nvm: https://github.com/creationix/nvm/blob/master/install.sh#L1

I would trust nvm as being a good citizen as for the shebang, has been here for a while now.

Daniel15:
See https://github.com/yarnpkg/yarn/pull/3338 for prior discussion on a similar change in Yarn itself.

vvo:
Thanks for that detailed link

badersur:
I didn't know that bash is not available on other OSes! And I thought you are using Bash from the `if` syntax.

Should I change the script to use `sh` shell instead? by
1. Changing the shebang to `#!/usr/bin/env sh`
2. Replacing `[[` with `[` for if statements

And modifying the installation doc to use `curl -o- -L https://yarnpkg.com/install.sh | sh`

badersur:
I've [found](https://stackoverflow.com/questions/11206233/what-does-the-operator-do-in-shell-scripts) that:
```sh
if [ ! "$REPLY" =~ ^[Yy]$ ]; then
    printf "$red> Aborting$reset\n"
    exit 1
fi
```
uses a bash syntax & I'm not sure how to convert it to `sh`!

badersur:
Theoretically, the script now uses `sh` syntax but I'm not sure about some code, such as:

```sh
if echo $version | grep -qE "^[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$"; then
```

and 
```sh
if [ $YARN_PROFILE = *"fish"* ]; then
```

Daniel15:
Thanks for continuing to work on this 😃 

> I'm not sure about some code, such as:
> 
> ```source-shell
> if echo $version | grep -qE "^[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+$"; then
> ```

This one is just `grep`, so it should be fine as long as the installed version of `grep` supports it.

> and
> 
> ```source-shell
> if [ $YARN_PROFILE = *"fish"* ]; then
> ```

I don't think this one will work: https://github.com/koalaman/shellcheck/wiki/SC2081

badersur:
You're welcome :)

I was just worried about the regex inside grep! I still don't know how to use regexes in shell scripting & I agree it should be fine :+1:

I liked the shellcheck project so I installed it & added the VSCode extension! I'm planning to implement their suggestions (I've found that my little `case` is incorrect :cry:)

Thanks for letting me know about this amazing tool :smiley: 

I think it'll be nice to add it to your Travis CI script :)

badersur:
I'm almost done :smiley: 

There're things I'm not sure about but the script is working fine :)

I'll explain my changes after some break :smile:

badersur:
I tried changing `printf` to use `"..%s.." "$var"` syntax but it looks ugly, not easy to read and colours' strings are printed on the terminal as is (without affecting the output with the colour): 

```sh
printf "%s> Version number must match MAJOR.MINOR.PATCH.%s\n" "$red" "$reset
```

Maybe my code is wrong but I couldn't achieve previous behaviour using `printf` (and I like to use `echo` for no obvious reason!)

I don't understand what's the value of `YARN_PROFILE="$(yarn_detect_profile)"` will be! The function doesn't return a value (and I don't understand the purpose of `local` & shellcheck warned me about it so I just commented it out :smile:)

Also, there's a warning about `${fish_user_paths}` which I don't know from where it'll be obtained (seems like an exported variable!)

No matter what I did, `sh` kept complaining about `read -p "$1 [y/N] " -n 1 -r` in the last function so I just converted it to `echo` and got rid of the headache :smile_cat: 

@Daniel15 I think the changes are complete but maybe testing from users of `fish` shell would be helpful to double check :)

badersur:
In [other PR](https://github.com/yarnpkg/website/issues/389), I've noticed that `YARN_PROFILE` is checked to be empty then printed to the terminal. I've used these steps to test:

I removed my write permission from profile files using:
```sh
cd ~; chmod u-w .bashrc .bash_profile .zshrc .profile
```

and changed the if statements to use `-w`, then yarn returned:
```
> Profile not found. Tried **YARN_PROFILE value missing** (as defined in $PROFILE), ~/.bashrc, ~/.bash_profile, ~/.zshrc, and ~/.profile.
> Create one of them and run this script again
> Create it (touch **... missing**) and run this script again
   OR
> Append the following lines to the correct file yourself:

export PATH="$HOME/.yarn/bin:$PATH"
```

which doesn't seem alright to me! Maybe we should change the message here and update the script to use `-w` in that PR.

badersur:
It turns out that [`echo` isn't portable](https://unix.stackexchange.com/tags/echo/info) :sob: and I don't know how to work with `printf` to colourise the output :frowning:

badersur:
Hello, Are there any news about this PR? :)

PR #550, opened 2017-06-27
feat(Detail): add link to yarn.fyi in sidebar

### LINKED ISSUE #549: Add short permalink to package info page
https://yarnpkg.com/en/package/react should display `https://yarn.fyi/react` somewhere, so that people know that it works as a nice short permalink to the page. 

@Haroenv do you have any design ideas for this? Perhaps we could put it above the home page link in the right column.

Haroenv:
closes #549

screen shot 2017-06-27 at 10 19 23

http://deploy-preview-550--yarnpkg.netlify.com/en/package/yarn


Haroenv:
Per @vvo's suggestion I added a copy button, but still a link inside that. This link stays on the same page, but links to yarn.fyi for consistency. 

screen shot 2017-06-27 at 15 41 50

vvo:
Copy must include protocol (https://) no? right now it does not

_sass/_details.scss
```diff
@@ -18,6 +18,25 @@
         width: 2em;
         opacity: .5;
       }
+
+      &__yarn {
+        display: flex;
+        align-items: center;
+
+        .copyable {
```
Daniel15: Is this styling actually specific to this use case, or could it be reused?

Daniel15:
This looks pretty good to me! Could you make it show up as `yarn.fyi/{name}` (no `https://`), but copy `https://yarn.fyi/{name}` when clicked? Showing the full URL including `https://` looks slightly cluttered IMO, but we definitely need to copy it like @vvo mentioned.

Daniel15:
Actually maybe `https://` is fine. I'll leave it up to you 😛

Haroenv:
I don't have a general purpose way of copying something that's not visible, it might be a little confusing that you get copied something else than what is visible .

Daniel15:
Sure. Feel free to merge this whenever you like. 😃

Haroenv:
Weird things happen when the character count is a lot (e.g. [this-package-has-a-very-long-name-if-you-will-install-it-you-are-very-strong-lol](http://deploy-preview-550--yarnpkg.netlify.com/en/package/this-package-has-a-very-long-name-if-you-will-install-it-you-are-very-strong-lol))

For some reason the truncation happens too late and the icon is moved to the side before truncating.

screen shot 2017-06-27 at 19 16 05

vvo:
https://clipboardjs.com/ has a way to copy something inside an attribute rather than the text.

vvo:
> For some reason the truncation happens too late and the icon is moved to the side before truncating.

force the width of the text content to %?

Daniel15:
> has a way to copy something inside an attribute rather than the text.

clipboardjs.com likely takes the same approach as we take at Facebook - Create a hidden element and copy text from it:
```js
let success;
const node = document.createElement('textarea');
document.body.appendChild(node);
node.value = text;
node.select();
try {
  success = document.execCommand('copy');
} catch (e) {
  success = false;
}
root.removeChild(node);
```

Haroenv:
I didn't know node.select() was now supported, will remember that in the future 😄

PR #551, opened 2017-06-27
fix(blog): don't check for language dropdown if not i18n

### LINKED ISSUE #546: Language selector missing on blog page
https://yarnpkg.com/blog/

Console throwing error regarding missing ID. This is due to the language dropdown missing at the top of the page.

Haroenv:
The blog page doesn't have a language dropdown, so this event listener adding does not make sense.

fixes #546 

You can check this by seeing if there are console errors on the blog page comparing [this preview](https://deploy-preview-551--yarnpkg.netlify.com/blog) with the [live version](https://yarnpkg.com/blog)


Haroenv:
I checked in the preview, it works now

PR #552, opened 2017-06-27
Support installing versions that start with v

badersur:
Maybe I'm the only one who try installing a specific version with the same version shown in the release page (i.e v0.27.0): `curl -o- -L https://yarnpkg.com/install.sh | sh -s -- --version v0.27.0`

But, I think it's nice to support both ways instead of showing the error:
> Version number must match MAJOR.MINOR.PATCH.

Edit:
It seems that there're other changes to be done to support this! Sorry, I didn't notice that you're using another variable for the version `specified_version=$2`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,8 +16,11 @@ yarn_get_tarball() {
   elif [ "$1" = '--rc' ]; then
     url=https://yarnpkg.com/latest-rc.tar.gz
   elif [ "$1" = '--version' ]; then
-    # Validate that the version matches MAJOR.MINOR.PATCH to avoid garbage-in/garbage-out behavior
     version=$2
+    if echo $version | grep -q "^[v]"; then
+      version=$(echo $version | cut -d "v" -f 2)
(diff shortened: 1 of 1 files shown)


Daniel15:
> It seems that there're other changes to be done to support this! Sorry, I didn't notice that you're using another variable for the version `specified_version=$2

Yeah, I think you're right. Perhaps the script could be cleaned up to remove that duplication. Definitely test it out and ensure it works :)

badersur:
I've tested these changes locally & it worked as expected (and I'm thinking about changing them to use an if else statements!). Then I saw that variable and panicked :smile: 

I'll need sometime to read the whole script to understand it...

badersur:
If the script supports this, does it mean that package managers should support it as well?!

After rethinking, it may be better to mention in the docs that we should use version *number*. Or just leave it as is and display the current message:
> Version number must match MAJOR.MINOR.PATCH.

PR #554, opened 2017-06-28
global: add docs for getting/setting symlink location

paulirish:
I learned something new in https://github.com/yarnpkg/yarn/pull/3458#issuecomment-310984999. :)  

I had looked for this information on this page many times before so this should some other people some time.


paulirish:
Thanks man!

bestander:
Thank you for contributing, very much appreciated

PR #564, opened 2017-07-08
fix(404): make Copyable back in line with change

Haroenv:
I changed how Copyable works in #550 but forgot that it was also used in this 404-ish page. This fixed that!


js/src/lib/Details/index.js
```diff
@@ -170,13 +178,21 @@ class Details extends Component {
             this.props.objectID
           )}
         </h2>
-        <p>{window.i18n.detail.not_found.yours}</p>
+        <p>
+          {window.i18n.detail.not_found.yours}
```
Haroenv: These are `prettier` changes, probably because this file hasn’t been changed since last version update or something.

PR #567, opened 2017-07-11
Add note about cmdtest in Linux installation docs

xbili:
This is quite a common gotcha for people trying to install `yarn` on Ubuntu 17.04. Would be nice if it was included in the documentation to provide an early heads up.

Refer to [this](https://github.com/yarnpkg/yarn/issues/2821) for more details.


BYK:
👍

PR #580, opened 2017-07-24
Workspaces blog post

bestander:
Started a blog post about Yarn Workspaces.

--- update 31.07.2017
For those waiting for this PR to get merged.
The team decided to hold off the blog post for a couple of weeks to allow some time for bug fixing.
Feedback is still welcome.


Haroenv:
Deploy preview ready!

Built with commit 85747b353125aded8f28bab4970a1177ba6f815e

https://deploy-preview-580--yarnpkg.netlify.com

Bnaya:
The file extension is not `.md`  so GH won't process the markdown for preview

_posts/2017-07-26-introducing-workspaces.md
```diff
@@ -0,0 +1,237 @@
...
+  "dependencies": {
+    "chalk": "^1.1.3",
+    "diff": "^3.2.0",
+    "jest-matcher-utils": "^20.0.3",
+    "pretty-format": "^20.0.3"
+  }
+}`
+```
+
+An approach that a wrapper like Lerna takes is to run run `yarn install` for each package.json separately and then run `yarn link` for packages that depend on each other.
```
robdonn: '... Lerna takes is to **_run run_** ...' ?

svenefftinge:
So the behavior is the same as using --hoist in lerna. Maybe you should mention that.

rarkins:
Is the `yarn.lock` content in the project root expected to be the same regardless of which package.json subdirectory it's run from?

_posts/2017-07-26-introducing-workspaces.md
```diff
@@ -0,0 +1,237 @@
...
+modified: ../../yarn.lock
+```
+
+Note that **workspaces** don't have yarn.lock and the **root** yarn.lock contains all the dependencies for all the **workspaces**.
+
+## Integrating with Lerna
+
+Do Yarn Workspaces make Lerna obsolete?
+
+Not at all.
```
thejameskyle: This might be a bit misleading. While the current implementation of Yarn Workspaces might not totally replace the functionality of Lerna, the intention is that it will.

Yarn should solve many of the same problems and people should start adapting their tools to work on top of Yarn Workspaces instead of Lerna. I want to eventually deprecate Lerna altogether.

bestander:
> Is the yarn.lock content in the project root expected to be the same regardless of which package.json subdirectory it's run from?

yarn.lock in the root will contain all the resolutions for all workspaces' package.json files.
If you run `yarn install` in a workspace folder Yarn will automatically cd to the root and run it from there

_posts/2017-07-26-introducing-workspaces.md
bestander: This could be a different discussion for a different time.
By the time Yarn arrives to feature parity with Lerna the latter may have some opinionated and specific features that Yarn team would not want to copy or maintain.

It will be up to Lerna maintainers to make this decision.

thejameskyle: I designed Lerna so that it intentionally wouldn't have features like that, things that were added should be entirely rethought on top of workspaces

PR #586, opened 2017-07-28
Make nightly installer oneliner consistent with other installation docs

f:
Easiest way was not the easiest way. It's now changed to a oneliner to make it easier and consistent with *install a specific version* document.

Ref: https://github.com/yarnpkg/website/blob/master/lang/en/docs/_installations/tarball.md#installation-script

### DIFF
--- a/lang/en/docs/_installations/nightly.md
+++ b/lang/en/docs/_installations/nightly.md
@@ -2,9 +2,7 @@
 
 The easiest way of installing a nightly build is via our shell script:
 ```sh
-wget https://yarnpkg.com/install.sh
-chmod +x install.sh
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit 9327f68161ea5ca2bbbfda8547141ebc8f13050e

https://deploy-preview-586--yarnpkg.netlify.com

f:
```
curl -o- -L https://yarnpkg.com/install.sh | sh /dev/stdin --nightly
```
is the other alternative not to use `bash` command for the servers not using `bash`.

Daniel15:
This looks good to me. Thanks!

PR #592, opened 2017-08-07
[WIP] Revamp download page

Daniel15:
The "Linux" tab was getting a bit unwieldy given there's now several unofficial packages alongside our official Debian and RPM packages. It seems like tabs are the wrong UI element for this page now that it has grown, so I've replaced the tabs with a dropdown list:

This was inspired by the Let's Encrypt / EFF Certbot site (https://certbot.eff.org/), which has a similar dropdown. Design input would be appreciated.

In addition, I've added instructions for installing the RC and nightly releases. Currently, installation of RCs is undocumented, and nightly releases have their own separate page at https://yarnpkg.com/en/docs/nightly.

For RC and nightly releases, the instructions are slightly different. To handle this, I wrapped the version-specific instructions in DIVs with particular classes (`install-only-stable` for stable instructions, `install-only-rc` for RCs, and `install-only-nightly` for nightlies) and show the right one depending on the selection.

**Test plan**
Go to http://localhost:4000/lang/en/docs/install/, it defaults to your current OS
Go to http://localhost:4000/lang/en/docs/install/#asdasdasdasd (invalid OS in URL), it defaults to your current OS
Go to http://localhost:4000/lang/en/docs/install/#debian (OS but no version in URL), it shows Debian instructions
Go to http://localhost:4000/lang/en/docs/install/#debian-rc (OS and version in URL), it shows Debian RC instructions
Switch OS in dropdown, verify that correct instructions are shown
Switch version in dropdown, verify correct instructions are shown


Haroenv:
The build failed: 

```
[...]
3:58:11 AM: Liquid Exception: Could not locate the included file '_installations/alpine.md' in any of ["/opt/build/repo/lang/zh-Hans/docs"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source. in /_layouts/pages/install.html
[...]
```

Daniel15:
Huh... Do I need to create the page for other languages? The English version (`lang/en/docs/_installations/alpine.md`) exists.

Haroenv:
Looking at it now, it works locally

Haroenv:
I can't build the production version myself, because downloading from crowdin gives me this message: 

```
error: Seems Crowdin server API URL is not valid. Please check the `base_url` parameter in the configuration file.
make[1]: *** [crowdin-download] Error 1
make: *** [build-production] Error 2
```

Haroenv:
Not sure what caused it exactly, but https://github.com/yarnpkg/website/commit/a7ba2e41cbdb45edea3e932a7fe2aae3902fba1a can be deployed

BYK:
>In addition, I've added instructions for installing the RC and nightly releases. 

<3

Haroenv:
I'm wondering if those pages should be made manually on crowdin, so that the download works 🤔

Daniel15:
Yeah, I have no idea. cc @thejameskyle - Do you remember if we have to do anything special when new pages are added?

thejameskyle:
Something should be running this script, but it doesn't appear like anything is: https://github.com/yarnpkg/website/blob/43fec17ba4f1378f18795607c39fdfcefa452748/Makefile#L31

Daniel15:
@thejameskyle Does that only run on master? Some of these .md files aren't in master yet, they're just in this PR.

thejameskyle:
Oh right... I think we may have never run into this scenario before (where something in crowdin is explicitly depending on something not in crowdin – normally files are totally independent of one another). You might just need to run that script manually.

Haroenv:
When I do `make crowdin-upload` I get messages about the URL being wrong (after I set the API key)

```sh
$ make crowdin-upload
error: Seems Crowdin server API URL is not valid. Please check the `base_url` parameter in the configuration file.
make: *** [crowdin-upload] Error 1
```

`base_url` doesn't exist in the `crowdin.yaml`, but it does have `base_path: .`, would that be related?

Daniel15:
I promise I haven't forgotten about this... I'll come back to it and figure out the Crowdin thing. 😛

mikemaccana:
PS. I *love* treating each Linux distro as an OS in it's own right, with top level native packages.

Daniel15:
Yeah... The "Linux" tab was getting very large, and there's not enough room to add many more tabs.

mikemaccana:
Sure but 90% are using Debian/Ubuntu or RHEL/Fedora. Always can do 'other' for users of yggdrasil etc.  😉

Haroenv:
@Daniel15, I wanted to try out the Crowdin upload script, but there are some conflict in the linux installations that I'm not sure which to keep, could you fix those conflicts? ❤️

Daniel15:
Sure, I'll try to take a look again soon.

Daniel15:
Confirmed it's working: https://deploy-preview-592--yarnpkg.netlify.com/en/docs/install

I'm going to merge this seeing as it's been sitting around for 8 months with no objections. We can iterate on the design and make it nicer :)

Haroenv:
How did you get it to work in the end @daniel15?

mikemaccana:
Just wanted to say I really like this design - it goes stright to the best way to deploy packages for your OS. Well done.

Daniel15:
> How did you get it to work in the end @Daniel15?

@Haroenv I ran the CrowdIn sync command on my computer, using the same access token as the Netlify build uses.

PR #593, opened 2017-08-09
feat(Search): allow to click on a keyword to refine it

Haroenv:
Hey, 

So what happens here, is that we add a refinementList which renders nothing, but is used for its logic alone. When someone clicks on a keyword, it gets refined and shown in `currentRefinements`, and when a refinement is removed from there or another widget, it gets unrefined too.

This was an experimentation to see if this pattern was possible (so for that, yes it works @mthuret, @kikobeats)

### DIFF
--- a/_sass/_search.scss
+++ b/_sass/_search.scss
@@ -282,9 +282,8 @@ $ais-muted-color: rgba(black,.5);
       border-bottom: dotted 1px;
     }
 
-    &:hover,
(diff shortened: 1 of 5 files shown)


Haroenv:
Deploy preview ready!

Built with commit d92301600d5a20a65844cd1472a283169e98030c

https://deploy-preview-593--yarnpkg.netlify.com

PR #602, opened 2017-08-17
feat(detail): add "tags" to the sidebar

### LINKED ISSUE #601: Show the version of other tags
The tags `beta` and `next` and how they relate to the version number could be showed somewhere near the version number or the Versions in the sidebar

Haroenv:
It's showing which tags you can install a package from, and what versions that corresponds to. 

I wanted to have something actionable (like copy the install script for that version, but I'm not sure if it makes sense. 

It won't show up on all packages yet, since there hasn't been a full reindex since, but one that has tags for example is `@rize/blocks`, one that doesn't have it is `react-instantsearch`. 

closes #601

### DIFF
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -344,6 +344,7 @@ script:
     devdependencies: DevDependencies
     packages: Packages
     see_package_json: see package.json
+    tags: Tags
     versions: Versions
     display_all: Display all
     hide: Hide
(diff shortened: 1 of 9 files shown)


Haroenv:
Deploy preview ready!

Built with commit 13a248601ba9fe50957bf7619c259af360f8ed94

https://deploy-preview-602--yarnpkg.netlify.com

Haroenv:
Wait, since we can only squash, I’ll do this in two PRs

arcanis:
Nice!

PR #632, opened 2017-09-07
Adds the workspace documentation

arcanis:


Haroenv:
Deploy preview failed.

Built with commit b9ccd5c930f35f8effc019ad89d69fd755d64b84

https://app.netlify.com/sites/yarnpkg/deploys/59b14b39cf321c5fc5d3fe06

lang/en/docs/workspaces.md
```diff
@@ -0,0 +1,92 @@
+---
+id: docs_workspaces
+guide: docs_workspaces
+layout: guide
+---
+
+{% include vars.html %}
+
+Workspace are a new way to setup your package architecture that's available by default starting from Yarn 1.0. It allows you to setup multiple packages in such a way that you only need to run `yarn install` once to install all of them in a single pass.
```
BYK: s/Workspace/Workspaces

BYK: >you only need to run `yarn install` once to install all of them in a single pass

you only need to run `yarn install` at the root, to install everything in a single pass.

lang/en/docs/workspaces.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Why would you want to do this?
+
+- Your dependencies can be linked together, which means that a workspace can depend on another and always use the most up-to-date version of it, without having to run `yarn install` again like with the `file:` protocol, and with better safety mechanisms than with `yarn link`.
```
BYK: Better to split into two sentences:

Your dependencies can be linked together, which means a workspace can depend on another and use the latest version of it from the monorepo. This is also a safer mechanism than `yarn link` since yarn now keeps track of actual dependencies along with versions even if they are linked locally.

lang/en/docs/workspaces.md
```diff
@@ -0,0 +1,92 @@
...
+
+And that's it! Requiring `workspace-a` from a file located in `workspace-b` will now use the exact code currently located inside your project rather than what is published on Github, and the `cross-env` package has been correctly deduped and put at the root of your project to be used by both `workspace-a` and `workspace-b`.
+
+### How does it compare to Lerna?
+
+Yarn's workspaces are the low-level primitives that tools like Lerna can use. In fact, [they actually do](https://github.com/lerna/lerna/pull/899)! Workspaces will never try to support the high-level feature that Lerna offers, but by implementing the core part of the resolution and linking part of the workspaces inside Yarn itself we hope to enable new usages and better performances.
+
+### Tips & Tricks
+
+  - The `workspaces` field is an array containing the paths to each workspace. Since it might be tedious to keep track of each of them, this field also accepts glob patterns!
```
BYK: May be suggest the packages/* pattern which is pretty common?

lang/en/docs/workspaces.md
arcanis: they can be installed from anywhere, actually, as long as you're in the workspace tree

lang/en/docs/workspaces.md
```diff
@@ -0,0 +1,96 @@
...
+/workspace-b/package.json
+```
+
+And that's it! Requiring `workspace-a` from a file located in `workspace-b` will now use the exact code currently located inside your project rather than what is published on Github, and the `cross-env` package has been correctly deduped and put at the root of your project to be used by both `workspace-a` and `workspace-b`.
+
+### How does it compare to Lerna?
+
+Yarn's workspaces are the low-level primitives that tools like Lerna can (and [do](https://github.com/lerna/lerna/pull/899)!) use. They will never try to support the high-level feature that Lerna offers, but by implementing the core logic of the resolution and linking steps inside Yarn itself we hope to enable new usages and improve performances.
+
+### Tips & Tricks
```
BYK: Just remembered: it is important to mention the following too:

 - Workspaces only works in private packages
 - You can disable workspaces via setting `workspaces-experimental` to `false` in `.yarnrc`

lang/en/docs/workspaces.md
```diff
@@ -0,0 +1,96 @@
...
+
+/workspace-a/package.json
+/workspace-b/package.json
+```
+
+And that's it! Requiring `workspace-a` from a file located in `workspace-b` will now use the exact code currently located inside your project rather than what is published on Github, and the `cross-env` package has been correctly deduped and put at the root of your project to be used by both `workspace-a` and `workspace-b`.
+
+### How does it compare to Lerna?
+
+Yarn's workspaces are the low-level primitives that tools like Lerna can (and [do](https://github.com/lerna/lerna/pull/899)!) use. They will never try to support the high-level feature that Lerna offers, but by implementing the core logic of the resolution and linking steps inside Yarn itself we hope to enable new usages and improve performances.
```
Haroenv: nit: `improve performance`

arcanis:
@Haroenv is there something to do for this page to appear on https://deploy-preview-632--yarnpkg.netlify.com/en/docs? It seems automatic, but I don't see it

Haroenv:
did you fill https://github.com/yarnpkg/website/blob/master/_data/guides.yml @arcanis

Haroenv:
yes, seems like that data file needs to be filled for it to render correctly, see https://deploy-preview-632--yarnpkg.netlify.com/en/docs/workspaces

lang/en/docs/workspaces.md
```diff
@@ -0,0 +1,102 @@
...
+
+  - The `workspaces` field is an array containing the paths to each workspace. Since it might be tedious to keep track of each of them, this field also accepts glob patterns! For example, Babel reference all of their packages through a single `packages/*` directive.
+
+  - Workspaces are stable enough to be used in large-scale applications and shouldn't change anything to the way the regular installs work, but if you think they're breaking something, you can disable them by adding the following line into your Yarnrc file:
+
+    ```
+    workspaces-experimental false
+    ```
+
+### Limitations & Caveheats
```
Haroenv: nit: `caveats`

raycohen:
The language 
> This is also a better mechanism than `yarn link` since it only affects your workspace tree rather than your whole system.
is a bit strong and could be interpreted as "`yarn link` should not be used". Is that intentional? I think a little rewording could help make it clear this benefit is useful when you are in a multi-package setup

netcazean:
#1108

PR #663, opened 2017-09-27
feat(search): focus on homepage

### LINKED ISSUE #662: Auto-focus front page search field
**Feature proposal:**

screen shot 2017-09-26 at 19 26 33

I propose that the main search field on the front page is focused automatically once the app is mounted, since it's such a prominent feature. It would make the frontpage immediately useful as a search page.

Haroenv:
fixes #662

What do you think @scarlac?

### DIFF
--- a/js/src/lib/Search/index.js
+++ b/js/src/lib/Search/index.js
@@ -13,7 +13,7 @@ const equals = (arr1, arr2) =>
 // package overview page
 // home page (/:lang/)
 const shouldFocus = path =>
-  path.includes("/packages") || path.split("/").length <= 3;
+  path.includes("/packages") || path.replace(/\/[a-zA-Z\-]+\/?/, '').length === 0;
 
 class RefinementList extends Component {
   componentWillReceiveProps(newProps) {
--- a/js/src/lib/Search/index.js
+++ b/js/src/lib/Search/index.js
@@ -13,7 +13,7 @@ const equals = (arr1, arr2) =>
 // package overview page
 // home page (/:lang/)
 const shouldFocus = path =>
-  path.includes("/packages") || path.split("/").length === 3;
+  path.includes("/packages") || path.split("/").length <= 3;
 
 class RefinementList extends Component {
   componentWillReceiveProps(newProps) {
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit 95af8af2fe5ef1dee1b2efa892294bcc4ecd037e

https://deploy-preview-663--yarnpkg.netlify.com

js/src/lib/Search/index.js
```diff
@@ -10,6 +10,11 @@ import { algolia } from '../config';
 const equals = (arr1, arr2) =>
   arr1.length === arr2.length && arr1.reduce((a, b, i) => a && arr2[i], true);
 
+// package overview page
+// home page (/:lang/)
+const shouldFocus = path =>
+  path.includes("/packages") || path.split("/").length === 3;
```
scarlac: Will we always have a trailing slash? As far as I can test, we always link to the version with a trailing slash but if I manually type "https://yarnpkg.com/en" it accepts the URL but the feature wouldn't work.

scarlac:
Awesome, @Haroenv ! Only 1 small comment (see commit). I see it activates on almost all pages via the menu which is fine for me although you asked if that was appropriate for the docs page :)

js/src/lib/Search/index.js
Haroenv: oh good point, let's change it to `<= 3`

js/src/lib/Search/index.js
```diff
@@ -10,6 +10,11 @@ import { algolia } from '../config';
 const equals = (arr1, arr2) =>
   arr1.length === arr2.length && arr1.reduce((a, b, i) => a && arr2[i], true);
 
+// package overview page
+// home page (/:lang/)
+const shouldFocus = path =>
+  path.includes("/packages") || path.split("/").length <= 3;
```
Haroenv: Change this so it works on `/en`, on `/en/`, not on `/en/docs` or `/en/docs/`

scarlac:
Tested it, and it works! Nice work, @Haroenv

PR #670, opened 2017-10-04
Don't use apt-key, don't blindly trust yarn for everything

KwadroNaut:
A tightly administered system wouldn't trust yarns apt key for any package. Administrated should still do the pinning though. Background: https://bugs.debian.org/861695


Haroenv:
Deploy preview ready!

Built with commit 56535ab6133fd9ffff380cc088c571a171a10103

https://deploy-preview-670--yarnpkg.netlify.com

xtuc:
I would even prefer a direct link to the `.deb` (or equivalent for other distros), what do you think?

KwadroNaut:
@xtuc super bad idea. If for whatever reason there needs to be an update to that package, people will not know they're out of date. Unless you have implemented a self updating package with TUF or something similar?

xtuc:
Yes, you're right. Nevermind.

Daniel15:
Hey, this is pretty clever. Does it work on older versions of Ubuntu and Debian? There's still many people using Ubuntu 14.04, unfortunately.

> I would even prefer a direct link to the .deb (or equivalent for other distros), what do you think?

It's not documented, but https://yarnpkg.com/latest.deb works.

PR #680, opened 2017-10-16
Allow link interaction in Read More sections

davidjb:
On a package pages with links in the readme (such as https://yarnpkg.com/en/package/ldap-promise), there are links visible but they are not clickable, even though the text is selectable with a mouse.  This CSS change prevents the pointer from interacting with the read more overlay, meaning the browser can interact with links and other UI elements.

The previous workaround was to open the read more section, but this is non-intuitive.

### DIFF
--- a/_sass/_readmore.scss
+++ b/_sass/_readmore.scss
@@ -12,6 +12,7 @@
       height: 250px;
       width: 100%;
       box-shadow: inset 0 -100px 60px -60px white;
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit d2a38a139ce4ec84f636ff0888217a8d846edec1

https://deploy-preview-680--yarnpkg.netlify.com

PR #686, opened 2017-10-20
fix(detail): make sure comments don't get flagged as email

### LINKED ISSUE #685: HTML Comments end up showing in package.html's readme section
I'm using a workflow which uses hidden annotations to link content across various markdown sources that compile into a single file before publishing and have been struggling with getting the same output to show correctly on yarn as it does on GitHub and NPM.

Initially we've use server-side-style-tags `<?…?>` which were 100% okay until we viewed the package on yarn.

We switched to rewrapping them before publishing like `<!--?…?-->` which works for us with some disadvantage but it works and it worked on yarn (with your current marked.js setup).

Then only enough we noticed that only one of those HTML-compliant comments is still lingering but is only partially-stripped, it does use a slightly different format than the rest which looks like `<!--@docs/….md@-->` in the MD file and outputs like `!--@docs/….md@--` on yarn (in Safari, and Chrome so far).

So one issue with two aspects in order of priority:

1. Can you please figure out why your setup treates `<!--?…?-->` and `<!--@docs/….md@-->` unequally, the first is stripped completely (which is great), the second is converted into and email link as the html:

```html
<a href="https://github.com/polyestr/mdon/master/mailto:!--@docs/README.md@--" title="null">!--@docs/README.md@--</a>
```

2. Can you please figure out how you could get marked to strip server-side-style-tags `<?…?>` only to the extent that other implementations (which might not be marked) use to render the same package README files (GitHub, NPM) have it done.

Thank you 👍

Haroenv:
fixes #685


Haroenv:
It's not an ideal fix, but the best I could find without forking marked

Haroenv:
Deploy preview ready!

Built with commit 89c1c8204cc21b2d052d670b707dc3dce93caf3e

https://deploy-preview-686--yarnpkg.netlify.com

Haroenv:
Deploy preview ready!

Built with commit c300e4fe81273e0fa3dd9fbaa1a030e4e107d2a1

https://deploy-preview-686--yarnpkg.netlify.com

Haroenv:
no idea why it's not working deployed 😢

PR #689, opened 2017-10-26
Feat/urlsync for keywords and owner

Haroenv:
cc @vvo

### DIFF
--- a/js/src/lib/Hit/index.js
+++ b/js/src/lib/Hit/index.js
@@ -20,8 +20,15 @@ export const Deprecated = ({ deprecated }) =>
     </span>
   ) : null;
 
-export const Owner = ({ link, avatar, name }) => (
-  <a className="ais-Hit--ownerLink" href={link}>
+export const Owner = ({ link, avatar, name, onClick }) => (
+  <a
+    className="ais-Hit--ownerLink"
+    href={link}
+    onClick={e => {
+      e.preventDefault();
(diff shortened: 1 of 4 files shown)


Haroenv:
Deploy preview ready!

Built with commit 08fc4dc406e1a48302e0486cdafabf4983985e31

https://deploy-preview-689--yarnpkg.netlify.com

vvo:
Not sure when is this triggered/where can I see it in action?

vvo:
I think I managed to use it but had a bit of trouble, because when searching for "vvo" first results are for packages I forked and the author is someone else..

Also the user link says it goes to github in a href but then it triggers a refinement instead.

Maybe we could just do a search for facet values query and show the matching user names on top of the search to allow to refine? And the link href would actually match the filtering url.

Then for the hits user, we could stick to GH link like we used to do.

It's a good feature but it's a bit confusing to use right now.

Haroenv:
I don't really understand what you mean, another searchbar or a refinementlist? There were some problems with wayy to many queries when having a virtual and real refinement list with the same data 🤔 

We can try getting the two together with the same list tho

vvo:
Let's sync on this next time we meet, for now that's a valid feature still :)

Haroenv:
Was aware of the drawbacks too, I mostly found it necessary to refine on a user from the url, and it would be weird if there was no interface to enable it.

PR #693, opened 2017-10-29
fix(details): manually sanitize relative svgs

### LINKED ISSUE #692: Broken images when using relative paths to SVG (compared to PNG)
*This is a joint issue for both NPM & Yarn caused by how `raw.githubusercontent.com` misidentifies certain file types.*

**TD;LR**

I created a sample package with a README and details:
  * **GitHub**: https://github.com/code-therapy/packages-with-readme-relative-images
  * **NPM**: https://www.npmjs.com/package/@code-therapy/packages-with-readme-relative-images
  * **Yarn**: https://yarnpkg.com/en/package/@code-therapy/packages-with-readme-relative-images

**Recommendations**

* Maybe GitHub can have a cross-origin policy for trusted referrers like npm & yarn.
* Maybe NPM can devise a way for packages to provide README with embedded content.
* Maybe you could use or borrow something from how rawgit handles this.

Thanks

---

NPM issue: npm/marky-markdown#420

Haroenv:
GitHub will return the wrong content-type and it will result in a broken image. This fix is only added in situations where String has `endsWith`, but that's okay since it's a very niche usecase

check: `@code-therapy/packages-with-readme-relative-images`

fixes #692 

cc @smotaal

### DIFF
--- a/js/src/lib/Details/Markdown.js
+++ b/js/src/lib/Details/Markdown.js
@@ -27,12 +27,16 @@ const renderAndEscapeMarkdown = ({ source, githubRepo }) => {
       });
 
     // manually ask for sanitation of svgs, otherwise it will have wrong content-type
-    const sanitize = href =>
-      `${href}${href.indexOf('//') === -1 &&
-      String.prototype.endsWith &&
-      href.endsWith('.svg')
-        ? '?sanitize=true'
-        : ''}`;
+    function sanitize(href) {
+      if (
+        href.indexOf('//') === -1 &&
+        String.prototype.endsWith &&
+        href.endsWith('.svg')
+      ) {
+        return `${href}?sanitize=true`;
+      }
+      return href;
+    }
 
     renderer.image = (href, title, text) =>
       `<img src="${prefix(
--- a/js/src/lib/Details/Markdown.js
+++ b/js/src/lib/Details/Markdown.js
@@ -26,8 +26,17 @@ const renderAndEscapeMarkdown = ({ source, githubRepo }) => {
         path,
       });
 
+    // manually ask for sanitation of svgs, otherwise it will have wrong content-type
+    const sanitize = href =>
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit 10fb4ff5512233012f9d96f715265935ba152b95

https://deploy-preview-693--yarnpkg.netlify.com

Haroenv:
screen shot 2017-10-29 at 07 38 18

I think the side effect when you _actually_ link raw.githubusercontent yourself is okay. Since that doesn't work on GitHub either so there should not be a reason to do it until they use the correct content type

js/src/lib/Details/Markdown.js
```diff
@@ -26,8 +26,19 @@ const renderAndEscapeMarkdown = ({ source, githubRepo }) => {
         path,
       });
 
+    // manually ask for sanitation of svgs, otherwise it will have wrong content-type
+    const sanitize = href =>
+      `${href}${href.indexOf('//') === -1 &&
+      String.prototype.endsWith &&
+      href.endsWith('.svg')
+        ? '?sanitize=true'
+        : ''}`;
```
Daniel15: This is very hard to read. Can you please make it a bit clearer? Also, shouldn't this only be done for relative links? Won't this current implementation add `?sanitize=true` for all SVG links, even if they're not GitHub links?

Daniel15: This is very hard to read, since it contains a multiline template string which contains a ternary. Can you please make it a bit clearer?  For example, something like:
```
function sanitize(href) {
  if (!href.endsWith('.svg')) {
    return href;
  }
  return href + '?sanitize=true';
}
```


Also, shouldn't this only be done for relative links? Won't this current implementation add `?sanitize=true` for all SVG links, even if they're not GitHub links?

Haroenv: it was already done only on relative links, made the conditional clearer now though

PR #695, opened 2017-10-30
fix(details): handle case where GitHub link 404s

Haroenv:
An example for this one is `duivvv` which has the "no readme data" from npm, then we fallback to GitHub, but the repo doesn't exist. It then throws an error because the readme is now `undefined`, which has no `length`. So we add a default to readme of an empty string


Haroenv:
Deploy preview ready!

Built with commit 61f775b31a257e26c167d803414ae90e8378fb09

https://deploy-preview-695--yarnpkg.netlify.com

PR #696, opened 2017-10-30
fix(search): don't cause rerenders of SearchBox component

Haroenv:
Because the referential equality of these two functions is different, it caused the SearchBox to rerender, and thus send a new query.


Haroenv:
Deploy preview ready!

Built with commit e3de007cd051c62ec05626a3223581bb354594c8

https://deploy-preview-696--yarnpkg.netlify.com

PR #702, opened 2017-11-03
feat(i18n): enable japanese

### LINKED ISSUE #701: "change language" icon
The icon itself reads "Aあ" since あ is Japanese hiragana, I supposed there should be Japanese translation, but it doesn't, this breaks the principle of least astonishment.

Haroenv:
AFAICT this should be all the changes needed, since it's 97% translated I think we should enable it. 

closes #701

### DIFF
--- a/_data/languages.yml
+++ b/_data/languages.yml
@@ -68,7 +68,7 @@
   accept_languages: ["it", "it-ch"]
 - name: 日本語
   tag: ja
-  enabled: No
+  enabled: Yes
   accept_languages: ["ja"]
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit d2ba10e03eede9f077312bc78f31aead1d237457

https://deploy-preview-702--yarnpkg.netlify.com

Haroenv:
Can't see it being broken anywhere except the few parts that aren't translated yet :)

Issue #721, opened 2017-11-17
yarn add <git remote url> doesn't work


kamleshchandnani:
```yarn add <git remote url>``` throws ```error Missing list of packages to add to your project``` because yarn is identifying the characters in git URL as paramter list and it breaks after ```@``` is encountered in the the git url.

Steps to reproduce:
```yarn add -d git+https://username@bitbucket.org/projectname/repo.git```

Possible solution:
```yarn add -d -- git+https://username@bitbucket.org/projectname/repo.git``` 
```--``` marks the end of parameter and then yarn installs the package successfully.

This is happening even if we try to use ```ssh```

**Comment by bestander** [comment]
The proper flag name is `--dev` or `-D`
And this works fine:
```
yarn add -D https://github.com/stevemao/left-pad
```

kamleshchandnani:
Avoid the typo
```yarn add -D -- git+https://username@bitbucket.org/projectname/repo.git```
This doesn’t works

bestander:
Can you give a specific example that does not work?
I don't have a bitbucket hosted package handy.
If you do it without `-D --` part it works then?

kamleshchandnani:
@bestander So basically what I mean is if your git repo has some special characters like '@' in this case ```yarn add -D -- git+https://username@bitbucket.org/projectname/repo.git```. If you remove ```--``` after your flag you won't be able to install the package. I have bitbucket repo but it wont work for you as its private to org

PR #722, opened 2017-11-17
Modified docs for yarn add command

kamleshchandnani:
Added changes about how to use `yarn add` when installing from git repositories.
This commit adds ```--``` command to install packages from git repositories.
Addressed #721

### DIFF
--- a/lang/en/docs/cli/add.md
+++ b/lang/en/docs/cli/add.md
@@ -44,8 +44,8 @@ You can also specify packages from different locations:
   haven't been published to the registry.
 3. `yarn add file:/path/to/local/tarball.tgz` installs a package from a gzipped
   tarball which could be used to share a package before publishing it.
-4. `yarn add <git remote url>` installs a package from a remote git repository.
-5. `yarn add <git remote url>#<branch/commit/tag>` installs a package from a remote
+4. `yarn add [options] -- <git remote url>` installs a package from a remote git repository.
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit f1c79843ffc5567c6cc28b26191e914e34a66c82

https://deploy-preview-722--yarnpkg.netlify.com

Daniel15:
This is a bit confusing because it doesn't explain what "[options]" means.

kamleshchandnani:
@Daniel15 The ```[options]``` means ```--dev/-D``` ```--peer/-P``` and all other options that are available when doing ```yarn add```. could you suggest a better option for these? or shall I add description of options too?

Haroenv:
Is that because the options _need_ to come before the repository?

kamleshchandnani:
@Haroenv So the whole command will be like ```yarn add -D -- <git remote url>```
So here
-D = options
-- = end of termination which is missing in the docs and I have raised the PR for the same.

kamleshchandnani:
@BYK any feedbacks?

bestander:
@kamleshchandnani thanks for your contribution.
I agree with the feedback above, passing flags does not need a `--` before the repo name.
I've commented in the linked issue.

I don't think we want to merge this PR as is.
Please feel free to correct me if I am wrong

Issue #723, opened 2017-11-20
add a CDNs header under use it


Haroenv:
* unpkg
* jsDelivr

<img width="383" alt="screen shot 2017-11-20 at 10 21 42" src="https://user-images.githubusercontent.com/6270048/33010945-a23cc09a-cddc-11e7-85b3-438fe4909116.png">

xtuc:
Here is also a CDN I maintain: https://bundle.run. It's a bundler a service + a CDN on front of it. Can we add it there?

Haroenv:
sure!

Daniel15:
Don't forget CDNJS, which is probably the largest one :)

Haroenv:
I don’t think CDNjs has a 1:1 npm <-> cdn mapping though, and neither are they putting something in package.json @Daniel15

Daniel15:
@Haroenv There's an Algolia index used on their site (https://cdnjs.com/). Could npm package name be indexed too? Their JSON files have an `npmName` field (see https://github.com/cdnjs/cdnjs/blob/master/documents/autoupdate.md)

Haroenv:
Good point, cdnjs _does_ have an Algolia index, so we could query that with the `npmName` as filter, or simply doing getObject (and dealing with double namings)

mjackson:
I'm a little concerned that the order of CDNs in this list is going to convey a preference to users. For example, on several of the packages that I maintain, like [react-router](https://yarnpkg.com/en/package/react-router), jsDelivr is currently at the top of the list of CDNs. However, we don't actually recommend that people use jsDelivr in our README.

Would it be feasible to do a quick grep through the README to see if it contains any references to a particular CDN before putting it in the list?

Haroenv:
That’s an interesting idea. We have to only make sure that it won’t shuffle the order once it’s already rendered, but the search can happen at indexing time to avoid that.

xtuc:
Just an idea, but what about a new field in the package.json? `CDNs` which are the ONLY CDNs that will show up on the page.

So that @mjackson can only show his unpkg and me only bundle.run :smile:

Daniel15:
> CDNs which are the ONLY CDNs that will show up on the page.

What's the use case though? If a file is available on multiple CDNs, why would some of them intentionally be hidden? I agree that being able to influence the order is good, but I don't understand the purpose of totally hiding particular options.

xtuc:
@Daniel15 ok, you're right.

[...]

Issue #614, opened 2017-08-29
Browse files doesn't seem to work


dguo:
Hi, I tried opening the file listing for https://yarnpkg.com/en/package/qs, and I got an error message: `Could not load file listing: Unexpected token < in JSON at position 0`. I got the same thing in incognito mode. Firefox gave me `Could not load file listing: JSON.parse: unexpected character at line 1 column 1 of the JSON data`, and Safari gave me `Could not load file listing: The string did not match the expected pattern.`. I tried other packages (webpack and left-pad), and they produced the same errors for me.

It looks like it's making a request to the GitHub API, and the JSON response looks okay, though I don't see any file info.

Daniel15:
Looks like a breaking change with the Unpkg site. We use their API to retrieve the file listing. I'll ping @mjackson about it.

https://github.com/unpkg/unpkg-website/issues/60

mjackson:
Sorry @dguo, my bad! Should be fixed now :)

dguo:
Thanks to both of you for the quick fix!

PR #725, opened 2017-11-21
Add CDNs header under Use it (#723)

MartinKolarik:
1. Adds "CDNs" (#723).
2. Replaces usage of unpkg.com API with jsDelivr API for the file browser as previously discussed, because it is more stable and built for production usage. The API is also versioned which means there are never breaking changes such as #614.

### DIFF
--- a/js/src/lib/Details/Cdn.js
+++ b/js/src/lib/Details/Cdn.js
@@ -1,0 +1,36 @@
+import React, { Component } from 'react';
+import { Di } from './';
+
+const JsDelivr = ({ name }) => (
+  <Di
+    title="jsDelivr"
+    description={
+      <a
+        href={`https://cdn.jsdelivr.net/npm/${name}/`}
+        target="_blank"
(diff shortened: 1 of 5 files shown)


Haroenv:
Deploy preview ready!

Built with commit a39dd0dccfada684757e0e2be7f3cddd09dc9ee7

https://deploy-preview-725--yarnpkg.netlify.com

js/src/lib/Details/Cdn.js
```diff
@@ -0,0 +1,36 @@
...
+  render() {
+    const {
+      name,
+    } = this.props;
+
+    return (
+        <article className="details-side--cdns">
+            <h1>{window.i18n.detail.cdns}</h1>
+            <dl>
+                <JsDelivr name={name} />
```
Daniel15: Just do `name={this.props.name}`, no advantage in spreading the object if it's only used in a single place.

MartinKolarik:
Split into #726 and #727.

PR #726, opened 2017-11-21
Add CDNs header under Use it (#723)

MartinKolarik:
As discussed in #723.

### DIFF
--- a/js/src/lib/Details/Cdn.js
+++ b/js/src/lib/Details/Cdn.js
@@ -1,0 +1,32 @@
+import React, { Component } from 'react';
+import { Di } from './';
+
+const JsDelivr = ({ name }) => (
+  <Di
+    title="jsDelivr"
+    description={
+      <a
+        href={`https://cdn.jsdelivr.net/npm/${name}/`}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        cdn.jsdelivr.net/npm/{name}/
+      </a>
+    }
+  />
+);
+
+class Cdn extends Component {
+  render() {
+    return (
+        <article className="details-side--cdns">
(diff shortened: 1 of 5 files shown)


Haroenv:
Deploy preview ready!

Built with commit 5ebf6f6a2e0a5770f953d37389d1ab4fa060514b

https://deploy-preview-726--yarnpkg.netlify.com

js/src/lib/Details/Cdn.js
```diff
@@ -0,0 +1,32 @@
...
+  />
+);
+
+class Cdn extends Component {
+  render() {
+    return (
+        <article className="details-side--cdns">
+            <h1>{window.i18n.detail.cdns}</h1>
+            <dl>
+                <JsDelivr name={this.props.name} />
```
Haroenv: Can you add the other ones from the issue as well?

js/src/lib/Details/Cdn.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+  />
+);
+
+const BundleRun = ({ name }) => (
+  <Di
+    title="bundle.run"
+    description={
+      <a
+        href={`https://bundle.run/${name}`}
```
xtuc: Could we add the current version of the package here? The template is `https://bundle.run/${name}@${version}`.

This will avoid to unconsciously always query the latest version of the package. We can probably apply this to other CDN as well.

MartinKolarik: Sure, we already know the latest version so I don't see why not.

Haroenv:
Let's iterate further on it to get cdnjs in there if we find a good way to do it.

PR #727, opened 2017-11-21
Replace unpkg.com API with jsDelivr API in file browser

MartinKolarik:
Replaces usage of unpkg.com API with jsDelivr API for the file browser as previously discussed, because it is more stable and built for production usage. The API is also versioned which means there are never breaking changes such as #614.

### DIFF
--- a/js/src/lib/Details/FileBrowser.js
+++ b/js/src/lib/Details/FileBrowser.js
@@ -5,10 +5,6 @@ import bytes from 'bytes';
 
 const SORT_ORDER = { directory: 1, file: 2 };
 
-function getBasename(path) {
-  return path.substr(path.lastIndexOf('/') + 1);
-}
-
 export default class FileBrowser extends React.PureComponent {
   state = {
     expandedDirs: {
@@ -28,8 +24,16 @@ export default class FileBrowser extends React.PureComponent {
 
   _fetchFiles() {
     this.setState({ error: null, files: null });
-    const url = this._getBaseURL() + '?meta';
+    const url = `https://data.jsdelivr.com/v1/package/npm/${this.props.objectID}@${this.props.version}`;
     fetch(url)
+      .then(response => {
+        if (!response.ok) {
+          return response.json().then(body => {
+            throw body;
+          });
+        }
+        return response;
+      })
       .then(response => response.json())
       .then(
         files => this.setState({ files }),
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit dbdb5495bd817452fb9c38ae97f3518fabdde688

https://deploy-preview-727--yarnpkg.netlify.com

js/src/lib/Details/FileBrowser.js
```diff
@@ -28,8 +24,16 @@ export default class FileBrowser extends React.PureComponent {
 
   _fetchFiles() {
     this.setState({ error: null, files: null });
-    const url = this._getBaseURL() + '?meta';
+    const url = `https://data.jsdelivr.com/v1/package/npm/${this.props.objectID}@${this.props.version}`;
     fetch(url)
+      .then(response => {
+        if (!response.ok) {
```
Daniel15: Shouldn't error handling be in ` catch` block, or as the second argument to `then`? Does jsdelivr not use proper HTTP codes for their responses?

js/src/lib/Details/FileBrowser.js
```diff
@@ -28,8 +24,16 @@ export default class FileBrowser extends React.PureComponent {
 
   _fetchFiles() {
     this.setState({ error: null, files: null });
-    const url = this._getBaseURL() + '?meta';
+    const url = `https://data.jsdelivr.com/v1/package/npm/${this.props.objectID}@${this.props.version}`;
```
Daniel15: Would prefer if this was pulled out into a separate function (eg. `getURLForPackageMetadata` or something like that)

js/src/lib/Details/FileBrowser.js
```diff
@@ -28,8 +24,16 @@ export default class FileBrowser extends React.PureComponent {
...
     fetch(url)
+      .then(response => {
+        if (!response.ok) {
+          return response.json().then(body => {
+            throw body;
```
Haroenv: I don’t understand why it’s thrown

js/src/lib/Details/FileBrowser.js
MartinKolarik: This was a surprise to me as well. `fetch()` [doesn't throw based on status code](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Checking_that_the_fetch_was_successful) but only if there was a network error. jsDelivr errors have the appropriate status code, and a message in the body. For a unified error handling, we need to parse the body and throw it as an error ourselves.

MartinKolarik:
@Haroenv there's modification time and sha256 hash but not sure if those two are useful here.

js/src/lib/Details/FileBrowser.js
MartinKolarik: See my comment above.

PR #731, opened 2017-11-28
Randomize order of CDNs

mjackson:
Randomizes the order of the CDNs in the CDN list on a package page. See discussion in #723

### DIFF
--- a/js/src/lib/Details/Cdn.js
+++ b/js/src/lib/Details/Cdn.js
@@ -47,16 +47,34 @@ const BundleRun = ({ name, version }) => (
 );
 
 class Cdn extends Component {
+  constructor(props) {
+    super(props);
+
+    this.cdns = {
+      jsdelivr: JsDelivr,
+      unpkg: Unpkg,
+      bundlerun: BundleRun,
+    };
+
+    this.order = Object.keys(this.cdns).sort(() => {
+      return Math.random() - 0.5;
+    });
+  }
+
   render() {
+    const items = this.order.map(key =>
+      React.createElement(this.cdns[key], {
+        key: key,
+        name: this.props.name,
+        version: this.props.version,
+      })
+    );
+
     return (
-        <article className="details-side--cdns">
-            <h1>{window.i18n.detail.cdns}</h1>
-            <dl>
-                <JsDelivr name={this.props.name} version={this.props.version} />
-                <Unpkg name={this.props.name} version={this.props.version} />
-                <BundleRun name={this.props.name} version={this.props.version} />
-            </dl>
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit 95382358e607d67591a1f315494da5a1e1d071b0

https://deploy-preview-731--yarnpkg.netlify.com

Haroenv:
They reorder on rerender now. Can you make the randomisation persist? 

Thanks, I think it makes sense ❤️

mjackson:
> They reorder on rerender now. Can you make the randomisation persist?

Ah, sorry :P done!

js/src/lib/Details/Cdn.js
```diff
@@ -47,16 +47,34 @@ const BundleRun = ({ name, version }) => (
...
+    };
+
+    this.order = Object.keys(this.cdns).sort(() => {
+      return Math.random() - 0.5;
+    });
+  }
+
   render() {
+    const items = this.order.map(key =>
+      React.createElement(this.cdns[key], {
```
xtuc: You could also assign it to a variable with a capital name and use it as a component.

Daniel15: ^ Yeah this is my preference, too

```
const items = this.order.map(key => {
 const CDNComponent = this.cdns[key];
  return (
    <CDNComponent
      key={key}
      ........
    />
  );
}
```

Daniel15:
This looks good to me!

I had another idea that requires much more effort, but would be nicer long-term: Display them in order of usage, per-package. For example, if most people load `babel-standalone` via unpkg, display unpkg first. I think this would satisfy your use case, @mjackson 

This would require jsdelivr, bundle.run and unpkg to have an API that provides usage statistics per package. jsdelivr has an pretty powerful API for this (https://github.com/jsdelivr/data.jsdelivr.com#get-package-usage-stats) but  but I don't know if bundle.run or unpkg collect or provide similar stats.

js/src/lib/Details/Cdn.js
mjackson: JSX compiles down to `createElement`. Don't be afraid to use the JavaScript API when you've got JavaScript variables, guys 😅

Haroenv:
Sorry for the confusion that this whole thing seems to be adding, I didn’t expect it. Thanks for participating @mjackson

js/src/lib/Details/Cdn.js
Daniel15: @mjackson - Consistency is good, though. Why use `React.createElement` at just one call site?

PR #733, opened 2017-11-29
fix(cdn): link to main file

Haroenv:
We want to link a usable file instead of a browse view, because we have the browse view already

Do you think this is a good idea @mjackson @Daniel15 ?

### DIFF
--- a/js/src/lib/Details/Cdn.js
+++ b/js/src/lib/Details/Cdn.js
@@ -6,7 +6,7 @@ const JsDelivr = ({ name, version }) => (
     title="jsDelivr"
     description={
       <a
-        href={`https://cdn.jsdelivr.net/npm/${name}@${version}/`}
+        href={`https://cdn.jsdelivr.net/npm/${name}@${version}`}
         target="_blank"
         rel="noopener noreferrer"
       >
@@ -21,7 +21,7 @@ const Unpkg = ({ name, version }) => (
     title="unpkg"
     description={
       <a
-        href={`https://unpkg.com/${name}@${version}/`}
+        href={`https://unpkg.com/${name}@${version}`}
         target="_blank"
         rel="noopener noreferrer"
       >


Haroenv:
Deploy preview ready!

Built with commit 50edfe0f7a7a7c277ad059dab512d4fa6b5b93b0

https://deploy-preview-733--yarnpkg.netlify.com

mjackson:
This is a really, really hard problem to solve. I've been trying to solve it for the past year, as a matter of fact :)

The problem is that there are no clear rules for publishing JavaScript today, so npm is crazy town. You never know what you're going to get in an npm package. Is it CommonJS? Is it a UMD bundle? Is it ready for the browser? Does it still need to be built by webpack/browserify? Who knows?! There are no rules.

The only rules we have right now are these:

- there's usually a `main` entry, which is probably CommonJS for most packages, but some packages that are specifically meant for the browser put a UMD build here (e.g. jquery)
  - if a package has no `main` entry, use `index.js`
- sometimes there's a `browser` field. This field was introduced by Browserify as a way to designate "browser-ready" versions of files in your package, but that still may need to be built by Browserify
- on rare occasions there is a `module` field. This was introduced recently by rollup as a way to indicate which files may contain ESM, but again, these may still need to be built by your bundler

What you may notice is missing from this list is some sort of standard that indicates whether a file is *actually* ready for the browser or not. It just doesn't exist! So currently unpkg tries really hard to guess which file to serve (since most unpkg requests are coming from browsers) but it's not that great because of the aforementioned issues.

Anyway, all of this is basically to say I would not recommend making this change. This is also why [I said that bundle.run](https://github.com/yarnpkg/website/issues/723#issuecomment-347324187) (or a similar auto-bundling service) is the only one that gives people a file that they can actually use in the browser!

I hope that makes sense :)

Haroenv:
Great points, I was aware of the annoying parts of automatically picking a file. However I thought you said that you thought the problem was that only bundle.run is a real file. 

Since this is not really the case, and it's not always ready for browsers, the overview might be the better choice.

❤️  thanks for your opinions and for your work on OSS

xtuc:
Yeah I agree, bundle.run is pretty cool :smile: 

What if we make two separated sections in the doc?
- Hosting services
- Bundling services

mjackson:
You can also read @Rich_Harris' comments on how bundle.run (packd) and unpkg are different [here](https://github.com/Rich-Harris/packd#is-this-like-unpkgcom)

PR #735, opened 2017-12-05
Added warning when a version other than specified is already present in the system , as a fix for issue #3807 from Yarn Repository

TiagoJoseMagalhaes:
Added a warning string in yellow indicating the present version and the specified version when the specified version is not the same as a present yarn version.

Currently there is no automated test plan for this change.
A manual change is possible however by change the present yarn version to something like 0.0.0.


Haroenv:
Deploy preview ready!

Built with commit 601f3256122b399be498fcfdc23e9fd0286bf144

https://deploy-preview-735--yarnpkg.netlify.com

Daniel15:
This seems reasonable to me!

PR #740, opened 2017-12-13
feat(detail): show bundled size of a library

Haroenv:
the default size is visible in Usage, and the gzip size on hovering. The link goes to bundlesize

closes https://github.com/algolia/npm-search/issues/91

cc @pastelsky. I added `yarn website; ${navigator.userAgent}` as the `User-Agent` header to the request.

cc @vvo


Haroenv:
Deploy preview ready!

Built with commit 613ae31eaa148393707b044c28d7e4e4e13816fa

https://deploy-preview-740--yarnpkg.netlify.com

js/src/lib/Details/Usage.js
```diff
@@ -66,6 +67,22 @@ const Usage = ({
...
+      {bundlesize && (
+        <Di
+          icon="download-size"
+          title={window.i18n.detail.bundlesize}
+          description={
+            <a
+              target="_blank"
+              rel="noopener noreferrer"
+              href={bundlesize.href}
+              title={`size: ${bundlesize.size}, gzip: ${bundlesize.gzip}`}
```
Daniel15: Should "size" be localized too?

js/src/lib/Details/index.js
```diff
@@ -118,17 +119,37 @@ class Details extends Component {
...
       });
     }
+
+    const { name, version } = this.state;
+
+    get({
+      url: `https://bundlephobia.com/api/size?package=${name}@${version}`,
+      type: 'json',
+      headers: {
+        'User-Agent': `yarn website; ${navigator.userAgent}`,
```
Daniel15: Does this actually work? I thought User-Agent was a "protected" header that could not be overwritten.

Daniel15: Ahh, never mind, TIL this changed at some point:

> Note: The User-Agent header is no longer forbidden, as per spec — see forbidden header name list (this was implemented in Firefox 43), so it can now be set in a Fetch Headers object, via XHR setRequestHeader(), etc.

(from https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name)

js/src/lib/Details/Usage.js
Haroenv: Oh good call

js/src/lib/Details/index.js
Haroenv: Yeah, it worked in my devtools

vvo: @Haroenv Is it an ok practice to do so? I would have created a dedicated "X-header" but WFM if that's completely OK.

vvo:
Currently failing for some reason:
https://deploy-preview-740--yarnpkg.netlify.com/en/package/algoliasearch

image

js/src/lib/Details/index.js
Haroenv: Changed it to `X-Bundlephobia-User`, cc @pastelsky

leonardosnt:
@Haroenv 

This feature introduced a new issue:



https://github.com/yarnpkg/website/blob/143ec0d3495800ccc8b1c84566e467af52d77cf4/js/src/lib/Details/Usage.js#L79-L80

it happens because `window.i18n.detail.bundlesize_text` (currently) does not exist in other translations and it was not
copied from default language because the default copy algorithm do a shallow copy. I think it can be fixed by doing a deeper copy here
https://github.com/yarnpkg/website/blob/2e0a866fbf2181874dd656f94d8ca1901b484231/_includes/i18n-as-script.html#L10-L15

How to reproduce:
- Select a different language (from English)
- Go to a page of any package
- E.g: https://yarnpkg.com/fr/package/debug
- Wait a bit. The package details will disappear and a error will be thrown.

Haroenv:
oh thanks @leonardosnt, that's a great catch. Do you want to do a PR to make the shallow copy a deeper copy? Maybe with JSON.stringify / parse?

leonardosnt:
I don't have all the setup here (gem, bundler...)  😢

Haroenv:
If you want, ir is also possible to just rely on the deploy preview

leonardosnt:
I think it can be done this way: 
```javascript
function copyDefaults(from, to) {
  for (var key in from) {
    if (from[key] !== null && typeof from[key] === 'object') {
      copyDefaults(from[key], to[key] || (to[key] = {}));
      continue;
    }
    if (to.hasOwnProperty(key) === false || to[key] === null) {
      to[key] = from[key];
    }
  }
}

copyDefaults(i18n_default, window.i18n);
```
See it working: https://repl.it/@leonardosnt/HairyImpureWaterstrider

Any concerns/thoughts about the code?

Haroenv:
It seems like that works @leonardosnt, feel free to make a PR ❤️

leonardosnt:
Here is: https://github.com/yarnpkg/website/pull/741

I hope everything works 😆

PR #744, opened 2017-12-27
Update 2016-11-24-offline-mirror.md

perlun:
I struggled with this for a while until I found where it was. Agreeably, running `yarn install` as root is semi-odd, but I'm currently working on a deployment pipeline where it totally makes sense. The _builders_ run everything as root inside a Docker container, the runtime Docker image runs the application inside another container as a non-privileged user.


Haroenv:
Deploy preview ready!

Built with commit 33f1393e759214c8d41bce7e2a29efdcf0a1cd51

https://deploy-preview-744--yarnpkg.netlify.com

PR #750, opened 2018-01-03
added global bin folder to PATH

bestander:
I prefer using install.sh script for managing Yarn on my Raspberry Pi.
This seems to work both for Mac and a Raspbian OS.

@Daniel15, does it look reasonable?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -81,7 +81,7 @@ yarn_verify_integrity() {
 yarn_link() {
   printf "$cyan> Adding to \$PATH...$reset\n"
   YARN_PROFILE="$(yarn_detect_profile)"
-  SOURCE_STR="\nexport PATH=\"\$HOME/.yarn/bin:\$PATH\"\n"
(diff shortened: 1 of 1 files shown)


Haroenv:
Deploy preview ready!

Built with commit 9a0ecb4556fb9e60137ba07db28b2b4f2a345e53

https://deploy-preview-750--yarnpkg.netlify.com

Haroenv:
I like it!

bestander:
All right, better sorry than safe.
Let's merge and see what happens.
I consider @Haroenv's response as a review :)

Daniel15:
This looks good to me, although I thought that Yarn was supposed to create symlinks from `~/.yarn/bin` to point to files in `~/.config/yarn/global/node_modules/.bin`

bestander:
Didn't really work when I was installing global libraries with Yarn in Raspbian.
Maybe we need to add ~.yarn/bin to PATH instead?

Issue #754, opened 2018-01-13
Show repository information for packages hosted on GitLab


mojavelinux:
If I publish a package to npm that's hosted on GitLab, the repository information (from the repository field in package.json) does not appear on yarnpkg.com.

Examples:

* https://yarnpkg.com/en/package/@antora/cli
* https://yarnpkg.com/en/package/semantic-release-gitlab

Yet Yarn claims to support GitLab repositories on the following page in the documentation.

https://yarnpkg.com/en/docs/package-json#toc-repository

Thus, my only option is to use the homepage field to point to the repository.

Please add support for repositories hosted on GitLab by showing the repository information on yarnpkg.com.

Haroenv:
Hey, thanks for asking for this support, we didn’t yet add that because nobody asked for the support on the website. Do you want to contribute to adding that data? 

It’s pretty odd that the readme doesn’t display for `@antora/cli`. I’d expect it to show, but maybe there’s a bug where it will always fetch GitHub? 

EDIT: no the bug is also in npm: registry.npmjs.com/semantic-release-gitlab (they use something closed-source for showing the readme on their website 🛩)

Relevant support should be added both to the website here for:

1. stars
2. repository activity
3. prefixing of links and images in readme

To [npm-search](https://github.com/algolia/npm-search) for: 

1. changelog fetching

mojavelinux:
Thanks for the quick response and for considering better support for packages on GitLab. GitLab is growing very rapidly in popularity, so I think this will be a welcomed enhancement. I'd be happy to help.

To clarify, I wasn't focused on the README. That's not the concern here. The concern is that the information about the repository is missing from the right-hand sidebar. Notice the difference between yarnpkg and npmjs.

* https://www.npmjs.com/package/@antora/cli
* https://yarnpkg.com/en/package/@antora/cli

There's no link to the repository on yarnpkg.

This is actually an opportunity for differentiation from npmjs.com. npmjs.com is currently showing a link which can't be navigated in the browser. yarnpkg could create a proper link to the repository on GitLab.

mojavelinux:
Here's the repository information in package.json.

```json
  "repository": "gitlab:antora/antora",
```

Haroenv:
Ah, that extraction is also done in `npm-search`, we can add another key `gitlabRepo` with all relevant information and display that on the site if it’s there

leonardosnt:
@Haroenv 

[...]

PR #763, opened 2018-01-25
Add support for Gitlab and Bitbucket.

leonardosnt:
See discussion in #754 

Example pages:
 - Package hosted on GitLab: https://deploy-preview-763--yarnpkg.netlify.com/en/package/semantic-release-gitlab
 - Package hosted on Bitbucket: https://deploy-preview-763--yarnpkg.netlify.com/en/package/@atlassian/aui


Haroenv:

Built with commit 2120ab8ec8e205e4e08d1afad25d71daa6ed0f69

https://deploy-preview-763--yarnpkg.netlify.com

js/src/lib/Details/index.js
```diff
@@ -102,54 +103,152 @@ class Details extends Component {
...
+        // So we need to 'translate' raw urls to api urls.
+        // E.g (https://gitlab.com/janslow/gitlab-fetch/raw/master/CHANGELOG.md) -> (https://gitlab.com/api/v4/projects/janslow%2Fgitlab-fetch/repository/files/CHANGELOG.md?ref=master)
+        // Once gitlab adds support, we can get rid of this workaround.
+        const apiUrl = `https://gitlab.com/api/v4/projects/${user}%2F${project}/repository/files/${encodeURIComponent(
+          filePath
+        )}?ref=${branch}`;
+        return get({
+          url: apiUrl,
+          type: 'json',
+        }).then(res => res.encoding === 'base64' && atob(res.content));
```
Haroenv: this seems like a bit much code in this one file, what about extracting it into a js file?

js/src/lib/util.js
```diff
@@ -144,16 +144,32 @@ function parseHighlightedAttribute({
   return elements;
 }
 
-export function packageJSONLink({ githubRepo }) {
-  if (githubRepo) {
-    const { user, project, path, head } = githubRepo;
+export function packageJSONLink({ repository }) {
+  if (!repository) {
```
Haroenv: WDYT if we linked to unpkg/jsdelivr? less logic and the direct file, + it will work without repository field

js/src/lib/util.js
```diff
@@ -245,3 +261,16 @@ inlineRenderer.paragraph = function(text) {
 export const safeMarkdown = input => ({
   __html: xss(marked(unescape(input), { renderer: inlineRenderer })),
 });
+
+export const i18nReplaceVars = (message, vars) =>
```
Haroenv: do you think you can reuse this in other places where we replace variables?

Haroenv:
By the way, when we land this I'd like to mention you on twitter, do you have an account?

leonardosnt:
> Do you have an example of a package with an unidentified repository? I couldn't find any

I used this to test: https://deploy-preview-763--yarnpkg.netlify.com/en/package/73 (As you can see, it's link to a Github user, not a repository)

> By the way, when we land this I'd like to mention you on twitter, do you have an account?

Yes, https://twitter.com/leonard0snt

js/src/lib/util.js
leonardosnt: Sure!

js/src/lib/util.js
leonardosnt: I think it would be better, I'll do this.

js/src/lib/Details/index.js
leonardosnt: You mean the `getRepositoryDetails` function?

js/src/lib/util.js
Haroenv: Feel free to open another PR after this one reusing this function 🙌

js/src/lib/Details/index.js
Haroenv: yes, I wonder if it can be extracted. Maybe relying too much on `this` though, instead it can be passed `repository` maybe?

leonardosnt: What should we do about the `setState`'s ? 
Should we use a reference to the Details component or wait until we fetch all information and return a new state (Promise)?

Haroenv:
👏👏

leonardosnt:
@Haroenv Hey, just wondering if you saw my question in the `getRepositoryDetails` stuff...

Haroenv:
Yes, the set state made it a bit complicated to extract, even though it’s possible if we have some external function, which will receive the “repository” data, and return an object which will be passed to setstate?

PR #775, opened 2018-02-13
fix(highlighting): make sure spaces in description highlight render as spaces

Haroenv:
The problem is that markdown will interpret a single space as identical to "nothing", whic it isn't.

So as a fix, in the code for safe highlighting of markdown, I wrote `xss(...) || ' '`, which will make sure that if nothing is returned by the markdown parser, I return a space. Technically it's possible now that in the case of a completely empty description, a single space is shown instead, but practically that can't happen, since npm will pick the beginning of the readme if the description is empty.

|before|after|
|---|---|
|screen shot 2018-02-13 at 11 00 48|screen shot 2018-02-13 at 10 57 09|

cc @samouss

### DIFF
--- a/js/src/lib/util.js
+++ b/js/src/lib/util.js
@@ -230,7 +230,7 @@ inlineRenderer.paragraph = function(text) {
 };
 
 export const safeMarkdown = input => ({
-  __html: xss(marked(unescape(input), { renderer: inlineRenderer })),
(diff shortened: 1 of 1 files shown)


Haroenv:

Built with commit b2508b9e1410c2507dec8df4d0486806bc142d4d

https://deploy-preview-775--yarnpkg.netlify.com

PR #781, opened 2018-02-19
Use curl instead of wget on CentOS / Fedora / RHEL

jornane:
The default install doesn't come with wget, but it does come with cURL. So this command has a better chance of working without requiring installing another dependency.


Haroenv:

Built with commit b32b5c2a905fac106aca08e947f407bd6dbbbf89

https://deploy-preview-781--yarnpkg.netlify.com

Daniel15:
Thanks!

Daniel15:
I'm surprised that it doesn't come with wget out-of-the-box!

PR #782, opened 2018-02-22
Add docs for `yarn install --verbose`

cvetanov:
Was searching for an option to print all of the info when running `yarn install` and found the exact option in an issue [#763](https://github.com/yarnpkg/yarn/issues/763) instead of having it in the docs, therefore I am adding it to the docs for future wanderers like me.

### DIFF
--- a/lang/en/docs/cli/index.md
+++ b/lang/en/docs/cli/index.md
@@ -50,3 +50,7 @@ can also specify an alternate port.
 --mutex network
 --mutex network:30330
 ```
+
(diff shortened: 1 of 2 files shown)


Haroenv:

Built with commit b3bd4b53e04f85be27d64afce8f459a13a1035c3

https://deploy-preview-782--yarnpkg.netlify.com

Haroenv:
If I understand that PR correctly, the flag is global, not just for install, can you also add it to the global flag list?

cvetanov:
Sure thing, here it is.

bestander:
Awesome, thanks @cvetanov!

Issue #794, opened 2018-03-18
[suggestion] don't open new tab for github links in package detail page


transitive-bullshit:
Yarn's package search is imho the best search engine for the npm registry that exists.

A very common scenario I find myself doing is opening up a number of search result detail pages in different tabs because I want to investigate different related modules and try to find the one I'd like to use. 99% of the time I want to open up the github repo to view the code directly for those results, as that's the quickest way at a glance to see how respectable a potential module is, so I inevitably end up clicking each of the github links on the N package detail pages which opens up N new tabs, then I have to go back and close each of the N yarn package detail pages.

That's not to say that the yarn detail page isn't useful -- it has extremely useful info on it.  I'm just proposing a very simple change to not open up the repo / github link in a new tab.

I'd be very surprised if I were the only one to find the current behavior of `target="_blank"` annoying, though I'm not sure how much visibility this issue will get on here.

Very happy to submit the couple line PR if you agree to make this minor change.

Thanks!

Haroenv:
Hey @transitive-bullshit, thanks for those nice words! I think that might make sense. They were originally added as target blank since there wasn't a detail page yet, but since that's there now, it might not be as obvious anymore to open in a new tab. 

I'll ask around to some other people if they have opinions about it, but it's probably a good idea (it's in Hit if you want to do a PR

PR #796, opened 2018-03-22
remove new tab from package detail links

transitive-bullshit:
addresses #794

### DIFF
--- a/js/src/lib/Details/Links.js
+++ b/js/src/lib/Details/Links.js
@@ -15,8 +15,6 @@ const images = {
 
 export const Link = ({ site, url, display, Tag = 'a' }) => (
   <Tag
-    target="_blank"
-    rel="noopener noreferrer"
     href={url}
     className={`details-links--link details-links--link__${site}`}
   >


Haroenv:

Built with commit 8a1b897072c5ccaf1e157e30488102810d13087e

https://deploy-preview-796--yarnpkg.netlify.com

Haroenv:
Did you intend to also change the target on the hits?

transitive-bullshit:
I'm not sure what you mean by hits, as I don't see the term used in this source.

As far as I can tell, all of the links on the package detail page which used to be new tabs should now be regular.

Haroenv:
Yes, but in your issue you were talking about the search results (hits), right?

transitive-bullshit:
Ahhh no, I was talking about the package detail pages which are already 1-level deep and having them `target=blank` at that point is pretty annoying.

Would you be open to considering this PR as-is and then we can discuss hit functionality in a separate issue?

transitive-bullshit:
Actually, I was just looking at it, and I can't see anywhere on the hits where `target=blank` is used.  So as far as I can tell, the hit links currently work as expected and it's just the package detail page that needs changing.

Haroenv:
Great! Sorry for the initial confusion

PR #812, opened 2018-04-25
fix(redirect): Redirect to English pages for unsupported langs

BYK:
**Summary**

Follow up to #811.

**Test plan**

Manual verification.

PR #813, opened 2018-04-25
chore: Remove obsolete index.html

BYK:
**Summary**

Removes the obsolete index redirect file after #812.

**Test plan**

Visit the main page with an unsupported languages set and makes sure it falls back to the English version.


BYK:
Just verified that this works as expected now, Should be good to merge.

Daniel15:
I think this was needed for dev, as the Netlify redirects don't work in dev
mode. Please test on your computer and ensure dev mode is OK.

BYK:
@Daniel15 is right, abandoning.

Daniel15:
Maybe we could add a comment to the file explaining why it's needed?

PR #823, opened 2018-05-23
Less bold security claims

vog:
The strong security claims convey a wrong impression of the actual state of the project:

* SHA-1 hashes are used even though SHA-1 is considered deprecated for quite a long time.
* Files are partly processed before the checksum is verified (namely, it is unpacked), violating security best practices and making Yarn unnecessarily vulnerable to exploits for tar, gzip, etc., as well as zip bombs.
* The number of volunteers is too small to react quickly on those issues: The SHA-1 issue is open for 1.5 years, the verification issue is open for 0.5 years.

This all isn't bad in itself. Yarn does most things right and is nevertheless a really strong competitor to NPM. But the website should be honest on that upfront, instead of giving a false sense of security.

### DIFF
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -37,10 +37,11 @@ homepage_feature_fast_description: |
   faster than ever.
 homepage_feature_fast_image_alt: Watercolour of cat riding a rocketship
 
-homepage_feature_secure_title: Mega Secure.
+homepage_feature_secure_title: Moderately Secure.
 homepage_feature_secure_description: |
-  Yarn uses checksums to verify the integrity of every installed package
-  before its code is executed.
+  Yarn intends to use strong cryptographic checksums to verify the integrity of every
+  installed package before any further processing.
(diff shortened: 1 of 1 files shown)


Haroenv:

Built with commit 01dee4133965fb55809def293117de2392aeae43

https://deploy-preview-823--yarnpkg.netlify.com

vog:
Thinking more about this, you might want to switch the order of items, such that "Reliable" is above "Security".

Daniel15:
> The SHA-1 issue is open for 1.5 years

I think that's being addressed in https://github.com/yarnpkg/yarn/pull/5042?

vog:
@Daniel15 I'm aware of this progress I'm glad that as least one of the issues will be fixed over the next days.

But there's still the unpacking issue. And it is not clear how fast future security issues will be fixed.

Which, again, isn't that bad for a volunteer project with limited resources. But it is a problem *for a project making bold security claims*.

So in the end, this just means that the website should have been adjusted 1.5 years ago. And should not be changed back until all security issues have really been solved.

PR #844, opened 2018-06-28
Guard against partial curl failures

dfellis:
I pinged the Twitter account with this earlier. The script was failing due to a status code 520 for the primary tarball. (This is apparently a nonstandard error indicating that the server returned a non-HTTP response, and apparently has happened with Cloudflare in the past, but that's all I could turn up on that.) See image:

screenshot_20180628_131110

The issue is that the curl command is getting two files at once, but only the first file failed and not both. It seems that curl does not return an error status code if the second file is fine.

This `--fail-early` argument I found in the man page and indicates it was added for exactly this sort of situation. It only exists in curl 7.52.0 and above, which is 1.5 years old. I'm not sure if that's old enough for your intended usage here, since curl errors out on unknown arguments.

This could be refactored into a bash variable conditional on the curl version detected, if you prefer.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -29,7 +29,7 @@ yarn_get_tarball() {
   fi
   # Get both the tarball and its GPG signature
   tarball_tmp=`mktemp -t yarn.tar.gz.XXXXXXXXXX`
-  if curl --fail -L -o "$tarball_tmp#1" "$url{,.asc}"; then
+  if curl --fail-early --fail -L -o "$tarball_tmp#1" "$url{,.asc}"; then
     yarn_verify_integrity $tarball_tmp
 
     printf "$cyan> Extracting to ~/.yarn...$reset\n"


Daniel15:
>  It only exists in curl 7.52.0 and above, which is 1.5 years old.

Debian Stable comes with 7.52, but do you know which version of cURL comes with Mac OS? MacOS tends to come with pretty old versions of bash, curl, etc. so we might need to check the version and only use this flag if the newer version of cURL is available.

dfellis:
I haven't used a mac for about 2 years, so I don't know the answer to that.

cURL is under the MIT license. I thought Apple only paused updates of GPL licensed OSS?

Daniel15:
cc @yarnpkg/core - Does anyone that uses MacOS know which version of cURL it comes with?

torifat:
@Daniel15 curl `v7.54.0` with macOS `10.13.5`

BYK:
I know it's not ideal but we can split this into two lines maybe? We'll pay the connection price twice but I think that's acceptable?

OmgImAlexis:
Guessing this is related. Travis seems to be having issues again installing yarn.

```
0.95s$ curl -o- -L https://yarnpkg.com/install.sh | bash
Installing Yarn!
/usr/local/bin/yarn
(node:3426) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
> 1.3.2 is already installed, Specified version: <HTML>
<HEAD>
<TITLE>Bad Request</TITLE>
</HEAD>
<BODY BGCOLOR="white" FGCOLOR="black">
<H1>Bad Request</H1>
<HR>
<FONT FACE="Helvetica,Arial"><B>
Description: Could not process this request. 
</B></FONT>
<HR>
</BODY>.
> Downloading tarball...
```

PR #868, opened 2018-10-02
Added jsDelivr hits to package details

Drgy:
Algolia provides jsDelivr hits in it's API, since the data are already there they only need to be displayed in package details.

### DIFF
--- a/js/src/lib/Details/Popularity.js
+++ b/js/src/lib/Details/Popularity.js
@@ -52,6 +52,28 @@ const Dependents = ({ dependents, humanDependents, name }) =>
     />
   );
 
+const formatHits = (hits) => {
+  if (hits >= 1e9) {
+    return (Math.round(hits / 1e7) / 100) + "B";
+  } else if (hits >= 1e6) {
+    return (Math.round(hits / 1e4) / 100) + "M";
+  } else if (hits >= 1000) {
+    return (Math.round(hits / 10) / 100) + "k";
+  }
+
+  return hits;
+};
+
+const JsDelivrHits = ({jsDelivrHits}) => (<Di
+  icon="dependents"
+  title={window.i18n.detail.jsdelivr_hits}
+  description={
+    <span title={jsDelivrHits.toLocaleString(window.i18n.active_language)}>
+      {formatHits(jsDelivrHits)}
+    </span>
+  }
+/>);
+
 class Popularity extends Component {
   render() {
     const {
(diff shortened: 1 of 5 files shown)


js/src/lib/Details/Popularity.js
```diff
@@ -76,6 +99,7 @@ class Popularity extends Component {
               humanDependents={humanDependents}
               name={name}
             />
+            <JsDelivrHits jsDelivrHits={jsDelivrHits} />
```
Haroenv: I think It would make sense to have that number (a) in the sidebar where the jsdelivr url is, or (b) in a new thing for usage, where jsdelivr is one of the keys and unpkg etc. also are available. Wdyt?

Haroenv: I don't think this metric is more important than the npm downloads, and now that I looked at it, last 30 days is probably too long to fit on that line. To me it looks good, but would just like the input of @mjackson of unpkg to see if he'd rather see unpkg on there too :)

MartinKolarik: >I don't think this metric is more important than the npm downloads

I meant below downloads, above dependents. You might be right about the text length issue though 😄

MartinKolarik: image

Haroenv: maybe it's easier to read if it's just `jsDelivr last 30 days`

Drgy:
@Haroenv I updated the text, let me know if you think I should change anything else.

js/src/lib/Details/Popularity.js
```diff
@@ -52,6 +52,28 @@ const Dependents = ({ dependents, humanDependents, name }) =>
...
+  }
+
+  return hits;
+};
+
+const JsDelivrHits = ({jsDelivrHits}) => (<Di
+  icon="downloads"
+  title={window.i18n.detail.jsdelivr_hits}
+  description={
+    <span title={jsDelivrHits.toLocaleString(window.i18n.active_language)}>
```
Daniel15: Does this work for all languages supposed by the Yarn site?

Daniel15:
Seems reasonable to me. What do you think, @Haroenv?

js/src/lib/Details/Popularity.js
Haroenv: yes, this worked and surprisingly for really old browsers too

Haroenv:
yep, was waiting to see if anyone had any reason why they were opposed.

PR #871, opened 2018-10-05
Update CLI commands comparison

joshuamabina:
yarn add is equivalent to npm install --save


Haroenv:
that's correct nowadays, thanks!

Haroenv:
Actually I thought your change was the other way around. Save is a default flag in npm, which is probably why it was removed. It's the same for me, this way maybe a bit more obvious :)

PR #873, opened 2018-10-14
How to add the install directory to PATH

doxanthropos:
The install directory is not added to the PATH automatically, so it would be nice to let the users know, they have to change that to be able to use the globally installed packages.
See: https://github.com/yarnpkg/yarn/issues/5353

### DIFF
--- a/lang/en/docs/cli/global.md
+++ b/lang/en/docs/cli/global.md
@@ -30,6 +30,15 @@ $ create-react-app
 
 `yarn global dir` will print the output of the global installation folder that houses the global `node_modules`. By default that will be: `~/.config/yarn/global`.
 
+### Adding the install location to your PATH
+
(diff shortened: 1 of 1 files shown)


Haroenv:

Built with commit a6b0d32a4393719c7bedfc1042b5a3db338fb95f

https://deploy-preview-873--yarnpkg.netlify.com

Daniel15:
Seems fine to me... Doesn't the installation page already say to do that though? Hmm.

doxanthropos:
@Daniel15 No. Maybe it is mentioned in another place of the documentation, but the installation page as well as the "global" page do not say anything about adding to the PATH.
If you think it should go to the installation page, I would recommend also putting at least a reference to it on the "global" page, because that is the point where people will stumble on it, when they try to use globally installed packages.

PR #875, opened 2018-10-30
Added back the user header for bundlephobia

pastelsky:
Hi @Haroenv . I'm guessing this header might have been removed because of CORS issues in the past. These have been fixed now. In the absence of an API key, this header is required to recognise the source of traffic. It would be great if we can put this back.

### DIFF
--- a/js/src/lib/Details/index.js
+++ b/js/src/lib/Details/index.js
@@ -253,6 +253,9 @@ class Details extends Component {
     get({
       url: `https://bundlephobia.com/api/size?package=${name}@${version}`,
       type: 'json',
+      headers: {	
+        'X-Bundlephobia-User': 'yarn website',	
+      },
     }).then(res =>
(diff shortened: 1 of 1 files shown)


Haroenv:
If I see it correct, it looks like you’re blocking traffic with this header. Is that on purpose?

pastelsky:
Traffic from yarnpkg.com was throttled and eventually blocked for a while as there was a high volume from bogus searches like - `hack-cheat-generator` type of queries. The system wouldn't scale up to handle all of those at that time. Now, we have sufficient safeguards in place so that it does not end up overwhelming the server.

pastelsky:
Is there anything else I can do to get this merged sooner?

Haroenv:
I’ve only been able to test this on my phone; there I didn’t see the bundle size so I am not sure it worked :)

PR #879, opened 2018-11-01
Add lifecycle methods and environment variables

mlunoe:
Adding section about lifecycle methods and environment variables available in these


Haroenv:
thanks!!

PR #884, opened 2018-11-15
Clarify when yarn.lock is or isn't updated

CodingItWrong:
Based on a conversation on this Twitter thread: https://twitter.com/arcanis/status/1063043732356898816

This clarifies the difference between when a bare `yarn install` uses `yarn.lock` and when it *updates* `yarn.lock`

I looked into making a change at https://yarnpkg.com/en/docs/installing-dependencies#toc-installing-dependencies but it seemed like that was too much to add for that introductory page. Linking to the `yarn install` page seemed sufficient for folks to dig into more detail and learn about these scenarios.

### DIFF
--- a/lang/en/docs/cli/install.md
+++ b/lang/en/docs/cli/install.md
@@ -25,6 +25,13 @@ If you need reproducible dependencies, which is usually the case with the contin
 Install all the dependencies listed within `package.json` in the local
 `node_modules` folder.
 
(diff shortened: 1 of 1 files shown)


Haroenv:
thanks! this is a good call. Might be good to have a dedicated page on the interaction with the lock file, what do you thin @arcanis ?

PR #887, opened 2018-11-16
Update debian.md, Add missing `sudo`

rashivkp:
Add missing `sudo`  in install command to avoid permission error.


Haroenv:
good catch!

PR #891, opened 2018-12-04
Allow latest and nightly versions under --version flag in install script

vikr01:
In environments like Netlify, [the `--version` flag is enforced when installing `yarn`](https://github.com/netlify/build-image/blob/master/run-build-functions.sh#L87). Allowing `latest` and `nightly` to get the latest version of `yarn` is a simple change that could allow anyone to easily use the latest versions of `yarn`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,9 +16,13 @@ yarn_get_tarball() {
   elif [ "$1" = '--rc' ]; then
     url=https://yarnpkg.com/latest-rc.tar.gz
   elif [ "$1" = '--version' ]; then
-    # Validate that the version matches MAJOR.MINOR.PATCH to avoid garbage-in/garbage-out behavior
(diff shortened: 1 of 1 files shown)


vikr01:
@Daniel15 For Netlify, [the `YARN_VERSION` environment variable has to be set](https://www.netlify.com/blog/2016/11/01/yarn-support-on-netlify/) or it [defaults to `1.3.2`](https://github.com/netlify/build-image/blob/master/run-build.sh#L23). It can be set either on Netlify's site, or through [`netlify.toml`](https://github.com/electron-react-boilerplate/site/blob/master/netlify.toml#L6).

I'm not sure if this sort of behavior exists on any other environments, which is why I think it's good for the install script to have it. I can PR Netlify otherwise.

Issue #848, opened 2018-07-12
Yarn version manager (yvm)


jakebolam:
Thoughts on adding links to using a yarn version manager? like https://yvm.js.org/

Daniel15:
Seems like an interesting idea. The installation script for yvm doesn't do any integrity checking (no SHA hash, no GPG signature) which is a bit concerning. I wonder if something like this should actually be built in to Yarn itself - I think there's already *some* support for using a different Yarn version per project, but it involves committing the yarn.js file to source control.

jakebolam:
Great point. Will start looking at the security thing (we're currently working on the version thing too). 

Yeah the idea was that by keeping it separate, yarn can do more forward, drastic changes, between versions (that may even output different lockfiles). Currently managing yarn across our 100+ repos (if we have to upgrade them all at once) is massively interruptful.

keego:
Any update on this since tophat/yvm#102 was merged and shipped in tophat/yvm#131?

Wondering if its safe enough to move to yvm (from a security standpoint) or if I should still recommend my team use the [scan-homebrew-prs-for-the-version-you-desire approach](https://stackoverflow.com/a/54130907/3234235).

jakebolam:
@keego the signature ensures that a mirror hasn't tampered with the yarn version

yvm installs have been steadyily increasing over the last 2 months (with around 85k installs a day now):
<img width="587" alt="screen shot 2019-02-12 at 1 05 32 pm" src="https://user-images.githubusercontent.com/3534236/52658310-b82ffb80-2ec8-11e9-8ea2-522c6c0a49cb.png">

arcanis:
I'm going to close this one as a dupe of #896 (it would be the opposite really, but most of the discussion is there so we might as well use the one with the most comments).

**tldr:** Even though I'm sure yvm is a great project, I would personally suggest using [`yarn policies set-version`](https://yarnpkg.com/en/docs/cli/policies#toc-policies-set-version) in order to version Yarn within a project. It's native, so it's one less dependency.

PR #895, opened 2018-12-18
Adds documentation for yarn policies

arcanis:
The following PR adds documentation for `yarn policies set-version` (https://github.com/yarnpkg/yarn/commit/158da6c6287cbc4fee900e3704f140c3391dc28d)

PR #896, opened 2018-12-24
docs: add docs on yarn version management

jakebolam:
Add docs for users who need to manage yarn versions per project.

Mentioned in https://github.com/yarnpkg/website/issues/848

### DIFF
--- a/_data/guides.yml
+++ b/_data/guides.yml
@@ -236,6 +236,13 @@
   - id: docs_workspaces
     path: /docs/workspaces
 
+- id: docs_yarn_version_management
+  title: docs_yarn_version_management_title
+  description: docs_yarn_version_management_description
+  pages:
+  - id: docs_yarn_version_management
+    path: /docs/yarn-version-management
+
 - id: yarn_organization
   title: yarn_organization_title
   description: yarn_organization_description
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -221,7 +221,7 @@ docs_workspaces_description: |
 docs_yarn_version_management: Yarn Version Management
 docs_yarn_version_management_title: Yarn Version Management
 docs_yarn_version_management_description: |
-  Managing multiple versions of yarn.
+  Manage your yarn version per project.
 
 yarn_organization_title: Yarn Organization
 yarn_organization_description: |
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -218,6 +218,11 @@ docs_workspaces_title: Workspaces
 docs_workspaces_description: |
   Link together your projects for easier maintenance.
 
+docs_yarn_version_management: Yarn Version Management
+docs_yarn_version_management_title: Yarn Version Management
+docs_yarn_version_management_description: |
+  Managing multiple versions of yarn.
+
 yarn_organization_title: Yarn Organization
 yarn_organization_description: |
   The Yarn organization is a collaboration of many companies and
--- a/lang/en/docs/yarn-version-management.md
+++ b/lang/en/docs/yarn-version-management.md
@@ -4,28 +4,18 @@ guide: docs_yarn_version_management
 layout: guide
 ---
 
-# yarn-versions-at-scale
-
 {% include vars.html %}
 
+## Motivation <a class="toc" id="toc-motivation" href="#toc-motivation"></a>
+As your usage of yarn grows, you will find yourself with projects that depend on certain versions of yarn.
+Once you hit this scale it can be difficult to keep all your projects compatible with the latest version of yarn.
+This becomes increasingly more difficult if you have many developers working on these projects, as it is near impossible to have all project-contributors running the same version of yarn.
+To mitigate this, you can manage your version of yarn per project.
 
-Using yarn with 100s of repos and 100s of developments
+#### Use `yvm`, the yarn version manager <a class="toc" id="toc-use-yvm-the-yarn-version-manager" href="#toc-use-yvm-the-yarn-version-manager"></a>
+You [can use yvm](https://yvm.js.org/) to manage your yarn versions
 
-Manually managing different yarn versions across projects is a pain. This fixes that.
+#### Commit the yarn binary to the project and use `yarn-path` in `.yvmrc` <a class="toc" id="toc-commit-the-yarn-binary-to-the-project-and-use-yarn-path-in-yvmrc" href="#toc-commit-the-yarn-binary-to-the-project-and-use-yarn-path-in-yvmrc"></a>
 
-Yeah the idea was that by keeping it separate, yarn can do more forward, drastic changes, between versions (that may even output different lockfiles). Currently managing yarn across multiple repos (if we have to upgrade them all at once) is massively interruptful.
-
-
-You can use
-### `yarn-path` <a class="toc" id="toc-yarn-path" href="#toc-yarn-path"></a>
-
-```shell
-yarn-path "./bin/yarn"
-```
-
-Instructs yarn to defer to another Yarn binary for execution. Useful if you want to bundle Yarn into your repository
-and have everyone use the same version for consistency. This was introduced in Yarn 1.0, so all developers must have Yarn >= 1.0
-installed.
-
-Value must be a relative file path, or `false` to disable (default).
+See [`yarn-path` in the yarnrc docs](/lang/en/docs/yarnrc/#toc-yarn-path)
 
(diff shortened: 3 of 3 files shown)


arcanis:
Sounds duplicate with https://github.com/yarnpkg/website/pull/895

Sounds like yvm also has little to offer compared to `yarn policies set-version` (wasn't actually aware of this tool before 😄), so we probably don't want to recommand it?

msrose:
There does seem to be some overlap between these docs and the `yarn policies` docs, but I do think yvm offers a viable alternative for those who don't want to commit yarn releases to their repositories. Familiarity with tools like nvm/rvm/pyenv/rbenv/etc. is another reason to consider something like yvm. 

Maybe augmenting the `yarn policies` docs with a link to yvm is a better option than a completely separate docs page, while making sure to include a note that yvm is a community tool and doesn't receive direct support from the yarn team?

arcanis:
> Curious to the motivation behind having the tool version itself?

Not relying on the network is a very common use case that's sometimes hard for people to figure out. Having an easy way to fill the need doesn't cost us much and bring a lot of value. Additionally, we *want* people to use the latest releases as much as possible (so that we don't get as many already-obsolete issues), so an easy upgrade path helps us maintaining Yarn.

Overall, I would avoid recommending third-parties on the Yarn documentation in general, and even more so when they are an alternative to something that's already supported. That would just bring confusion and possible security risks (we have no control on yvm).

That said, maybe the solution to your problem would be to bring to Yarn what yvm does right. Maybe it'll be out of scope for Yarn, but we can certainly discuss it in the code repository 😉

jakebolam:
Thanks for all the feedback @arcanis 

> Overall, I would avoid recommending third-parties on the Yarn documentation in general, and even more so when they are an alternative to something that's already supported. That would just bring confusion and possible security risks (we have no control on yvm).
Understandable not to support 3rd party libraries officially. We are open to giving the yarn core team write access to the yvm project. 


Having both options is great (not everyone is super keen on committing binaries to their projects, and yvm solves this)

> Not relying on the network is a very common use case that's sometimes hard for people to figure out. Having an easy way to fill the need doesn't cost us much and bring a lot of value. Additionally, we _want_ people to use the latest releases as much as possible (so that we don't get as many already-obsolete issues), so an easy upgrade path helps us maintaining Yarn.

I do feel yvm has a similar level of offline support, and an easy upgrade path as well. We should all strive to run the latest version of yarn.

Bessonov:
Disclaimer: I'm yarn newbie and I'm not sure I want to stick with it.

I think this is off topic, but both ways are not optimal. Or I'm missing something.

Assume fresh workspace with existing project:
```bash
apt-get update > /dev/null && apt-get install -y curl unzip > /dev/null
echo "v10.15.3" > .nvmrc
echo 'yarn-path ".yarn/releases/yarn-1.12.3.js"' > .yarnrc
echo "1.12.3" > .yvmrc
```

Installation of node/npm with nvm:
```bash
curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | bash
. /root/.bashrc
nvm install
Found '/tmp/test/.nvmrc' with version <v10.15.3>
Downloading and installing node v10.15.3...
```
As you can see, if `.nvmrc` exist, I just run `nvm install` to get desired version. Check.

I'm not sure that this both are very intuitive:
```bash
curl -o- -L https://yarnpkg.com/install.sh | bash
> Successfully installed Yarn 1.13.0!

. /root/.bashrc
yarn --version
internal/modules/cjs/loader.js:584
    throw err;
    ^

Error: Cannot find module '/tmp/test/.yarn/releases/yarn-1.12.3.js'
```
Why someone want to have `yarn-1.12.3.js` in repository? It's like to commit extracted `maven` binary... No way.

Of course I can look at `.yarnrc` and take appropriate actions. But it doesn't feels right. Furthermore, yarn policies breaks `yvm`.

And similar
```bash
curl -fsSL https://raw.githubusercontent.com/tophat/yvm/master/scripts/install.sh | INSTALL_VERSION="v2.4.0" bash
. /root/.bashrc
yvm install
You have provided an invalid version number. use "yvm ls-remote" to see valid versions.
Error: Invalid version number provided
```

But unexpected for me:
```bash
yvm use    
Installing yarn v1.12.3 in /root/.yvm
```

> Not relying on the network is a very common use case that's sometimes hard for people to figure out. 

Sorry, but I don't care about it. I have everything on multiple laptops, CI cache, proxies and even packaged into development (yes, I do) and build images for Jenkins. I prefer to have a clean git history.

> That said, maybe the solution to your problem would be to bring to Yarn what yvm does right.

A huge +1 to fix yarn instead of 3rd party tools. And yes, for me `nvm`, `sdkman`, `yvm` and friends are workarounds, not solutions.

arcanis:
> Why someone want to have yarn-1.12.3.js in repository? It's like to commit extracted maven binary... No way.

While I understand where you're coming from, reconsider this "no way" and put in perspective the advantages versus the cons.

Pros:
- No need to actually download Yarn
- No need to make sure it's the right version of Yarn
- "I prefer to have a clean git history" - this gives you the best git history - you can checkout any single commit from your history and be 100% that you will use the Yarn version you configured at the time.

Cons:
- It takes some space on the repository.

Do you prefer your application to work without hassle in a consistent way by default, or to spare a few KB in your repository?

Bessonov:
> you can checkout any single commit from your history and be 100% that you will use the Yarn version you configured at the time.

What about node_modules? This is more important than yarn or like (o:

Your pros and cons are all valid, that's not the question. And for every pro and con there is an answer. It's even not about to save ~4.5MB on every yarn update. But more about to have clean log, diff while debugging, merging, rebasing, code review and tooling like gitlab, gerrit and so on. I'm sure, that some of this tools supports excludes, but 1. not all tools and 2. I don't want to spend time to figure it out.

It's more about weights between pros and cons. And unfortunately, it's not enough to have it in repository. You still need to run commands to get it into PATH.

What's about something like:
```
echo "v1.12.3" > .yarnrc
curl -o- -L https://yarnpkg.com/install.sh | bash
> Found '/tmp/test/.yarnrc' with version <v1.12.3>
> .yarn/releases/yarn-1.12.3.js not found
> Download yarn-1.12.3.js
> Successfully installed Yarn 1.12.3!

yarn --version
> v1.12.3

yarn install
> Found '/tmp/test/.yarnrc' with version <v1.12.3>
> .yarn/releases/yarn-1.12.3.js found, nothing to do!
```

So you can decide to add yarn to repository or to ignore it.

arcanis:
> What about node_modules? This is more important than yarn or like (o:

[It's funny you mention that 🤡](https://twitter.com/arcanis/status/1080810479398854656)

Now you must think I'm some kind of mad man - but really, consider what's the real reason is for not committing your node_modules. Why do we commit our sources but not our third-parties? You answered it before: VCS simply wouldn't support it. Upgrading a dependency would generate huge diffs with literally thousands of files moving around, and thousand others being diff'd.

But if you instead check-in exactly one file per package you use (the compressed package), and if you put them in a flat directory structure, the equation becomes very different - VCS are actually pretty good at this! We actually use this feature in the [Yarn v2](https://github.com/yarnpkg/berry/tree/master/.yarn/cache) repository, and I encourage you to clone the repository and see by yourself whether the increased time in noticeable.

We have some improvements to do (we generate a few extra symlinks I'd like to remove once Node gets to implement its Loaders API), but this should give you an idea of the vision I pursue.

> But more about to have clean log, diff while debugging, merging, rebasing, code review and tooling like gitlab, gerrit and so on.

From my experience (I also integrate Yarn within the FB infra), checking-in Yarn in our repository was entirely beneficial, including from a tooling and history perspective. I have complete control over which version everyone is using, problems are solved with a simple "rebase on master please", and the upgrades can actually be reviewed precisely because they are in the history.

arcanis:
> It's more about weights between pros and cons. And unfortunately, it's not enough to have it in repository. You still need to run commands to get it into PATH.

Note that you need to have *a* Yarn in the PATH - any version starting from 1.0 is enough (that's august 2017).

Bessonov:
You try to solve problems, which I never had :)

Bessonov:
I'm using `yvm` for a while now and I'm pretty happy with it. What's about to add a community page and mention projects such as `yvm`?

> Note that you need to have a Yarn in the PATH - any version starting from 1.0 is enough (that's august 2017).

No, thanks. For some time I use [this way](https://github.com/Bessonov/dev-image/) to work. Well, it's still not perfect, but it saves me from a lot of burden.

PR #908, opened 2019-01-24
Update docs or issue on yarn workspaces?

leimonio:
Tried the above example locally with `yarn v1.13.0`.
In the root folder as expected `/node_modules/workspace-a` is added, which is a dependency on `workspace-b`,.
Is it reasonable to also have `/node_modules/workspace-b` even if it is not a dependency in any workspace?

### DIFF
--- a/lang/en/docs/workspaces.md
+++ b/lang/en/docs/workspaces.md
@@ -69,6 +69,7 @@ Finally, run `yarn install` somewhere, ideally inside the workspace root. If eve
 /node_modules
 /node_modules/cross-env
 /node_modules/workspace-a -> /workspace-a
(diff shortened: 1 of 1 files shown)


leimonio:
@arcanis Thanks for the fast reply on that! The thing is that it's not always clear when to expect a dependency added in the shared or the dedicated workspace `node_modules/`. But as far as I can understand this is something that changes inside the algorithm that sets the workspaces. It's good for now, but it'd be good to have a more explicit description of what to be expected on `yarn install`.

arcanis:
Basically for each package you are allowed to access everything listed in your dependencies+devDependencies. Nothing less, nothing more 🙂

Issue #913, opened 2019-01-27
`brew install yarn --without-node` no longer possible


rdmurphy:
[As of six days ago](https://github.com/Homebrew/homebrew-core/pull/36222) Yarn's recommended install method via Homebrew has been removed because Homebrew is [doing away with options in all core formulae](https://github.com/Homebrew/homebrew-core/issues/31510). (I kinda get the rationale, but, uh... yanking out the carpet from under everyone.)

Obviously there's a bigger issue here (what is the new recommended way to install Yarn on MacOS?) but currently that means what's in the docs is no longer valid. 😬

Haroenv:
That’s unfortunate! I think in the meantime the regular fallback installation method listed on “Linux” is a good alternative. Looks like the home brew thread isn’t fully resolved for now and anyone interested in the yarn use case can chime in there. Cc @daniel15

Daniel15:
I'd be happy with someone modifying the site to show the shell script above the Homebrew instructions in the MacOS installation instructions. The RC and Nightly instructions already do that, so you can just copy what they're doing (eg. see https://yarnpkg.com/en/docs/install#mac-rc)

brendanfalkowski:
Literally every time I have to use `brew` it blows up in my face. The curl method worked for me and (hopefully) associates properly to the `nvm`-selected version of Node.

fab1o:
Try: `brew install yarn --ignore-dependencies`

brendanfalkowski:
Just for a sanity check, if I installed Yarn last week using: `curl -o- -L https://yarnpkg.com/install.sh | bash`

And today I ran: `brew install yarn --ignore-dependencies`

Do I now have two instances installed?

Running `which yarn` returns `/Users/brendan/.yarn/bin/yarn`. Should I expect to see something different from the Homebrew installation?

rdmurphy:
@brendanfalkowski You probably do — `brew` will install `yarn` in a different place that's standard for `brew` installs.

What happens when you run `which -a yarn`? I *think* you'll see both of them. (I think you'll see the Homebrew one at `/usr/local/bin/yarn`.)

dorokei:
I ran: ``brew install yarn --ignore-dependencies``

Installing yarn without node was successful.
But ``brew doctor`` showed below warning.

```
Warning: Some installed formulae are missing dependencies.
You should `brew install` the missing dependencies:
  brew install node

Run `brew missing` for more details.
```

[...]

PR #916, opened 2019-02-05
Remove outdated reference to --without-node argument

rypit:
Homebrew/homebrew-core#36222 removed support for passing arguments to `brew install yarn`. As such, we can no longer specify the `--without-node` option.

### DIFF
--- a/lang/en/docs/_installations/mac.md
+++ b/lang/en/docs/_installations/mac.md
@@ -8,10 +8,10 @@ This will also install Node.js if it is not already installed.
 brew install yarn
 ```
 
-If you use [nvm](https://github.com/creationix/nvm) or similar, you should exclude installing Node.js so that nvm's version of Node.js is used.
+If you use [nvm](https://github.com/creationix/nvm) or similar, you can exclude installing Node.js so that nvm's version of Node.js is used by ignoring dependencies.
 
 ```sh
-brew install yarn --without-node
+brew install yarn --ignore-dependencies
 ```
 
 #### MacPorts


rypit:
Adresses #913 and yarnpkg/yarn#6946

Daniel15:
Maybe this should say something like "If you use [nvm](https://github.com/creationix/nvm) or similar, you should install Yarn using the bash script to avoid installing an extra version of Node.js" and include the tarball instructions (same as what we display for RC / nightly releases)?

Ideally we'd like if someone could create a Homebrew "tap" that includes the `--without-node` option.

rypit:
@Daniel15 Based on the various threads on this, it seems like users with nvm/nodenv could get the same benefit  of getting yarn without the homebrew version of node by passing the `--ignore-dependencies` flag, since `node` is the [only dependency for the formula](https://github.com/Homebrew/homebrew-core/blob/a34c721586fece6d36d9b60fec5760e1b3361f11/Formula/yarn.rb#L10). 

I've updated this PR to include the tweak to the notes for those users.

I'm happy to create a formula, but would that be a larger discussion since it would introduce the following concerns(?):
* Would we want to maintain a tap with an almost duplicate Formula that's not part of homebrew-core?
* Would we want to remove yarn from homebrew-core to avoid duplication?
* Where would the new formula live?

Daniel15:
>     * Would we want to remove yarn from homebrew-core to avoid duplication?

I think so... Depends on if the Homebrew team let us do that.

>     * Where would the new formula live?

It could probably live in here: https://github.com/yarnpkg/releases

Daniel15:
Anyways, this site change looks fine for now, so let's merge it.

rypit:
🙇Thanks for the merge @Daniel15!  

> I'm not sure if that's the best approach or not, since I'm not very familiar with Homebrew ... It could probably live in here: https://github.com/yarnpkg/releases

If we put it there, we'd have to add the repo url when telling mac users how to tap. Install instructions would look something like:

```bash
brew tap yarnpkg/yarn https://github.com/yarnpkg/releases.git
brew install yarnpkg/yarn --without-node # or something similar to that
```

Homebrew convention is that tap repos are named in the following format: `yarnpkg/homebrew-yarn`. This would allow us to `brew tap yarnpgk/yarn` which is a little less verbose.

I'm not sure either of the above are ideal. 

As long as the dependencies for yarn stay stable, the current formula in homebrew-core may be the best experience for homebrew users (no custom taps required, principal of least surprise, homebrew users will maintain it). 

If it's decided otherwise please let me know, I'm happy to get involved.

PR #921, opened 2019-02-20
GitHub -> npm

chrisvasz:
Is that supposed to say "npm" instead of "GitHub"? In the second example, I don't know how yarn would know what to download from GitHub in the case of differing versions. If I'm just missing something and "GitHub" is correct, then I'm happy to add some clarifying comments to this PR to help readers understand how yarn can know what to download from GitHub.


Daniel15:
cc @arcanis

arcanis:
Lgtm, thanks!

PR #927, opened 2019-03-01
Add --update-new to Semaphore instructions

lesleh:
Adding `--update-new` ensures that the Semaphore build updates the package lists before attempting to install Yarn. Otherwise it may fail to install if the package list doesn't contain the version you're attempting to install.


Haroenv:
Which documentation do you have for this change?

lesleh:
@Haroenv I got it in a support ticket I sent to them, but the [installer tool is documented here](http://semaphoreci.com/docs/how-to-install-dependency.html) along with the `--install-new` flag, and specificially mentions that it's useful for after adding a new apt source. It's faster than running `apt update` as it only updates newly added sources.

Haroenv:
perfect, thanks

PR #931, opened 2019-03-10
Pointing to the updated doc on workspace

areai51:
Some of the steps here are outdated and may not be needed. Hence adding a quote to point to the updated doc on this.


Haroenv:
thanks! cc @arcanis

PR #935, opened 2019-03-18
Add Gentoo Linux install instructions

Bigfootjon:
Yarn is available on Gentoo's portage package manager (1), and I
have a pending patch to yarn to provide correct install instructions
from within the yarn cli (2).

(1) https://packages.gentoo.org/packages/sys-apps/yarn
(2) https://github.com/yarnpkg/yarn/pull/7123

### DIFF
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -270,6 +270,7 @@ install_os_alpine: Alpine
 install_os_arch: Arch Linux
 install_os_centos: CentOS / Fedora / RHEL
 install_os_debian: Debian / Ubuntu
+install_os_gentoo: Gentoo Linux
 install_os_mac: macOS
 install_os_opensuse: openSUSE
 install_os_solus: Solus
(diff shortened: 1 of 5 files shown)


Haroenv:
looks like we will need a fallback for languages not yet translated here too, like for install-ci

Bigfootjon:
@Haroenv are you referring to the `file_exists` tag? (added in https://github.com/yarnpkg/website/commit/fe50d77f15a90dbb27be13b7edb6ee59ef1a402b)

Haroenv:
Yes, it's a bit of a hacky workaround, but is needed because of the i18n implementation

Bigfootjon:
Well the build succeeds now but the deploy preview doesn't show any install instructions. @Haroenv Is this expected?

Haroenv:
That's not expected, and likely an error in the (fairly hacky) workaround for missing files. Unfortunately I don't really have time to look into this _right now_

Bigfootjon:
I was able to fix the missing install documentation. Turned out to be a flaw with the implementation of  `file_exists` where it didn't strip trailing whitespace, I've inserted a commit to fix that but if you would rather I can split it out into a separate PR **or** I can rewrite my changes to match the style of the other use of `file_exists` (i.e. without whitespace around the file path)

Haroenv:
Thanks for the effort you put in this!ç

Bigfootjon:
Happy to help!

PR #938, opened 2019-03-25
Fix logical mistake

sonicdoe:
`^1.12.0` matches the latest _minor_ release instead of the latest _patch_ release. For example, it currently matches `1.15.2`.


arcanis:
Indeed 👍

PR #942, opened 2019-04-07
Remove reference to Homebrew --ignore-dependencies option

### LINKED ISSUE #941: Homebrew option `--ignore-dependencies` is unsupported
The Yarn documentation says to install Yarn via Homebrew like so:

```
brew install yarn --ignore-dependencies
```

It looks like this stemmed from #913 / #916.

I found that when using `brew upgrade` later on, dependencies for `yarn` were not ignored, causing `node` to be installed via Homebrew. I opened an issue with Homebrew about this (Homebrew/brew#5967), and found that the `--ignore-dependencies` was meant to be used by Homebrew developers only, and is not a supported flag.

Homebrew has updated their documentation to point this out (Homebrew/brew#5970), so the Yarn documentation should probably not instruct users to use this flag anymore. 

I'm not sure what the documentation should say instead, though. At least in my case, having `nvm` _and_ Node via Homebrew [doesn't cause errors anymore](https://github.com/Homebrew/brew/issues/5967#issuecomment-479073467), I just have an extra copy of Node installed that will never be used.

klanchman:
Resolves #941

The `--ignore-dependencies` option is meant for Homebrew developers, and is not supported.

### DIFF
--- a/lang/en/docs/_installations/mac.md
+++ b/lang/en/docs/_installations/mac.md
@@ -8,11 +8,7 @@ This will also install Node.js if it is not already installed.
 brew install yarn
 ```
 
-If you use [nvm](https://github.com/creationix/nvm) or similar, you can exclude installing Node.js so that nvm's version of Node.js is used by ignoring dependencies.
-
-```sh
-brew install yarn --ignore-dependencies
-```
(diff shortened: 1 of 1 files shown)


Haroenv:
thank you!

MikeMcQuaid:
Thanks @klanchman and @Haroenv ❤️

PR #943, opened 2019-04-09
add prepare on certain script names.

KAMEDAkyosuke:
I notice that `prepare` was certain script names. 

https://github.com/yarnpkg/yarn/blob/8a671f10c86b66baa74ef6f4cd24d4ca8ee5067a/src/cli/commands/install.js#L1200


Haroenv:
that's correct!

PR #960, opened 2019-06-02
Add info how to use PowerShell to add exclusions

yooakim:
I have added some more details on how to whitelist the yarn process/folders from default anti-virus in Windows.


Haroenv:
@dan15, is this correct?

PR #963, opened 2019-06-04
Using Yarn in Codefresh CI/CD pipelines.

kostis-codefresh:


Daniel15:
The formatting of the first line looks wrong (it's showing raw Markdown), and also the tabs no wrap to another line :(



See http://deploy-preview-963--yarnpkg.netlify.com/en/docs/install-ci

I wonder if we need to rethink the layout of this page, and perhaps have a dropdown list of all the CI systems, similar to the download page.

kostis-codefresh:
@Daniel15 I just used gitlab as an example. And I just noticed that gitlab has the same issue in live docs actually. 
gitlab

Do you have an idea on why gitlab works like this. I might then fix the issue for both Codefresh and gitlab.

Regarding tabs, I expect many more CI/CD systems will be added in the future. Is having two rows of tabs problematic?

lang/en/docs/_ci/codefresh.md
```diff
@@ -0,0 +1,42 @@
+[Codefresh pipelines](https://codefresh.io/docs/docs/configure-ci-cd-pipeline/introduction-to-codefresh-pipelines/) use Docker images in all their steps so it is very easy to use any [Yarn version](https://hub.docker.com/_/node/) in any pipeline.
```
Haroenv: I think I might be fixed by starting with an empty line here

kostis-codefresh: @Haroenv I tried adding an empty line. But there is a pre-commit hook that runs cleanups on js, md and scess files and removes it automatically :-| 
Any other ideas?

Haroenv: empty line with a comment in? Sorry for the confusion here, it's definitely my fault

kostis-codefresh: @Haroenv yep that did it! I also fixed the gitlab page with the same comment.

Haroenv:
Cool, this works well. But as @Daniel15 pointed out, it overflows now. Maybe we should add a scroll bar there?

kostis-codefresh:
@Haroenv well I can imagine several CI/CD systems will be added in the future.  Daniel suggested a dropdown. Who can take a final decision on this matter?

Anyway I am open to anything. I am more interested in the content itself. Feel free to make edits directly with the desired outcome.

Haroenv:
let's leave it like this indeed, if it creates more friction we can fix it maybe with a dropdown

PR #965, opened 2019-06-07
docs: add workspaces caveat note

josefaidt:
## Description

Added note regarding Workspaces using local dependencies if available rather than reaching for the registry. This addition was prompted by [a tweet](https://twitter.com/arcanis/status/1136385324161097730?s=20) from @arcanis as I believe it is a bit easier to understand relative to the existing mention of this caveat.

Additional note added as a nested list item under the existing mention. I've transcribed the tweet to fit in the context, however I would appreciate any feedback you may have.


josefaidt:
image

Perhaps this may not be the best way to implement this note due to the spacing below nested list items.

Issue #793, opened 2018-03-16
[docs] Document `link:` dependency types


rally25rs:
Yarn supports a `"otherPackage": "link:../otherPackage"` dependency type. This is not something that `npm` has, so we can't assume users will know how it works.

We should document this package type somewhere, since it is specific to Yarn.

Haroenv:
https://github.com/yarnpkg/website/blob/master/lang/en/docs/package-json.md is where I'd put it :)

PR #969, opened 2019-06-18
Document `link:` dependency type (#793)

p12tic:
This PR adds minimal documentation of the `link:` dependency type. The corresponding RFC is [0000-link-dependency-type](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-link-dependency-type.md) which has been merged in this PR: https://github.com/yarnpkg/rfcs/pull/34. The original implementation has been merged to yarn two years ago: https://github.com/yarnpkg/yarn/pull/3359.

The comment by @Haroenv on #793 suggested to put the documentation to https://github.com/yarnpkg/website/blob/master/lang/en/docs/package-json.md, but since there's no existing documentation of other dependency types like `file:` on that page, I decided to go the minimal route and just update the documentation of `yarn add` which already has several examples. I can update package-json docs in this or followup PR if we'd like to have this information in that page too.

### DIFF
--- a/lang/en/docs/cli/add.md
+++ b/lang/en/docs/cli/add.md
@@ -44,10 +44,13 @@ You can also specify packages from different locations:
     haven't been published to the registry.
 3.  `yarn add file:/path/to/local/tarball.tgz` installs a package from a gzipped
     tarball which could be used to share a package before publishing it.
-4.  `yarn add <git remote url>` installs a package from a remote git repository.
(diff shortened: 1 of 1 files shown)


alxtz:
This is nice! I've been trying to fix https://github.com/yarnpkg/yarn/issues/2165 and found this is actually the best fix. Documenting this is very helpful
https://deploy-preview-969--yarnpkg.netlify.com/en/docs/cli/add#toc-adding-dependencies

PR #970, opened 2019-06-21
Add more information to the resolutions example

StephanBijzitter:
Closes https://github.com/yarnpkg/yarn/issues/5759

I had serious issues trying to write something up that's short yet understandable.
To me, this explanation makes sense, but honestly... it could probably do with some improvement so feel free to suggest them, no hard feelings!

### DIFF
--- a/lang/en/docs/selective-version-resolutions.md
+++ b/lang/en/docs/selective-version-resolutions.md
@@ -34,11 +34,20 @@ Add a `resolutions` field to your `package.json` file and define your version ov
     "d2": "file:../d2-1"
   },
   "resolutions": {
-    "d2/left-pad": "1.1.1",
-    "c/**/left-pad": "1.1.2"
+    "d2/left-pad": "2.0.0",
+    "c/**/left-pad": "3.0.0",
+    "**/right-pad": "1.0.0"
   }
 }
 ```
+
+In this example, all packages requiring `right-pad` will be given the same version. However, this will not always work for your codebase. Any package that wants to require version 6.0.0, will be forced to require version 1.0.0 and that may result in packages breaking.
+
+Another approach is to scope the resolution to a specific dependency, allowing multiple versions to still be used. In this case, when you import `left-pad` directly, it will be version 1.0.0, but when `d2` imports `left-pad` directly it will get version 2.0.0.
+
+However, if `d2` imports a dependency and that dependency imports `left-pad`, it will not likely be version 2.0.0, as the resolution does not apply in this case. To prevent this from happening, `c` uses `**` to apply the resultion for all of its dependencies as well.
+
+Note that not one of these is the better resolution, as different use-cases require different resolutions.
 
 Then run `yarn install`.
 


lang/en/docs/selective-version-resolutions.md
```diff
@@ -34,12 +34,21 @@ Add a `resolutions` field to your `package.json` file and define your version ov
...
+    "**/right-pad": "1.0.0"
   }
 }
 ```
 
+In this example, all packages requiring `right-pad` will be given the same version. However, this will not always work for your codebase. Any package that wants to require version 6.0.0, will be forced to require version 1.0.0 and that may result in packages breaking.
+
+Another approach is to scope the resolution to a specific dependency, allowing multiple versions to still be used. In this case, when you import `left-pad` directly, it will be version 1.0.0, but when `d2` imports `left-pad` directly it will get version 2.0.0.
+
+However, if `d2` imports a dependency and that dependency imports `left-pad`, it will not likely be version 2.0.0, as the resolution does not apply in this case. To prevent this from happening, `c` uses `**` to apply the resultion for all of its dependencies as well.
```
Haroenv: is `c` the name of a directory here?

StephanBijzitter: Yeah those names also confused me a bit, but `c` is defined as a dependency: `"c": "file:../c-1"`

I chose not to change the original example (keeping this PR as small as possible), but it may be clearer to use actual dependency names here. `c` and `d2` could just `jquery` and `lodash`, which would make it easier to read I think.

Haroenv: I'd prefer this too, although for the `file` dependency, something like `my-local-dependency` might stop some confusion too

StephanBijzitter:
>Just to see if I understand correctly. You mean if d2 imports a dependency, c, which depends on leftpad. Then the resolution c/**/left-pad would apply?

@shui91, that's an interesting case... I have not tested that. I would assume that's indeed what will happen, meaning `d2` resolution to have a single version of `left-pad` would result in two versions of `left-pad` because `c` also enforces its own version of `left-pad`.

Will test this sometime later today as this got my very curious.

rokumatsumoto:
@StephanBijzitter any update on this?

StephanBijzitter:
Alright, just tested.

The structure of the test is as follows, where `->` means "has a dependency on":

test-project -> dep-a -> dep-b -> dep-c -> jquery@3.1.1
test-project -> jquery@^3.4.1

For test-project, the package.json is as  follows:
```
  "dependencies": {
    "dep-a": "../dep-a",
    "jquery": "^3.4.1"
  }
```

And the lockfile is as follows:
```
jquery@3.1.1:
  version "3.1.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.1.1.tgz#347c1c21c7e004115e0a4da32cece041fad3c8a3"
  integrity sha1-NHwcIcfgBBFeCk2jLOzgQfrTyKM=

jquery@^3.4.1:
  version "3.4.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.4.1.tgz#714f1f8d9dde4bdfa55764ba37ef214630d80ef2"
  integrity sha512-36+AdBzCL+y6qjw5Tx7HgzeGCzC81MDDgaUP8ld2zhx58HdqXGoBd+tHdrBMiyjGQs0Hxs/MLZTu/eHNJJuWPw==
```

After adding the following resolution:
```
  "dependencies": {
    "dep-a": "../dep-a",
    "jquery": "^3.4.1"
  },
  "resolutions": {
    "dep-a/**/jquery": "3.3.1"
  }
```

The lock file changes to:
```
jquery@3.1.1, jquery@3.3.1:
  version "3.3.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.3.1.tgz#958ce29e81c9790f31be7792df5d4d95fc57fbca"
  integrity sha512-Ubldcmxp5np52/ENotGxlLe6aGMvmF4R8S6tZjsP6Knsaxd/xp3Zrh50cG93lR6nPXyUFwzN3ZSOQI0wRJNdGg==

jquery@^3.4.1:
  version "3.4.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.4.1.tgz#714f1f8d9dde4bdfa55764ba37ef214630d80ef2"
  integrity sha512-36+AdBzCL+y6qjw5Tx7HgzeGCzC81MDDgaUP8ld2zhx58HdqXGoBd+tHdrBMiyjGQs0Hxs/MLZTu/eHNJJuWPw==
```

So dep-c, which requires 3.1.1 now gets 3.3.1, while test-project which requires ^3.4.1 still gets 3.4.1.

That all sounds good, until we change the `dep-a/*/jquery` into `dep-a/*/jquery`. I would've expected this change to mean "only enforce this on direct dependencies of dep-a (which would be dep-b), but instead something weird happened:

```
jquery@3.1.1:
  version "3.1.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.1.1.tgz#347c1c21c7e004115e0a4da32cece041fad3c8a3"
  integrity sha1-NHwcIcfgBBFeCk2jLOzgQfrTyKM=

jquery@3.3.1:
  version "3.3.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.3.1.tgz#958ce29e81c9790f31be7792df5d4d95fc57fbca"
  integrity sha512-Ubldcmxp5np52/ENotGxlLe6aGMvmF4R8S6tZjsP6Knsaxd/xp3Zrh50cG93lR6nPXyUFwzN3ZSOQI0wRJNdGg==

jquery@^3.4.1:
  version "3.4.1"
  resolved "https://registry.yarnpkg.com/jquery/-/jquery-3.4.1.tgz#714f1f8d9dde4bdfa55764ba37ef214630d80ef2"
  integrity sha512-36+AdBzCL+y6qjw5Tx7HgzeGCzC81MDDgaUP8ld2zhx58HdqXGoBd+tHdrBMiyjGQs0Hxs/MLZTu/eHNJJuWPw==
```

So as for your question: yes, that is exactly what will happen.
And as for this weird behaviour, I'll create a separate bug report for that.

lang/en/docs/selective-version-resolutions.md
StephanBijzitter: Completely re-wrote it. It's a bit lengthy now, but with easier names and easier version numbers, hopefully it's relatively easy to read.

I also removed the file dependencies. I felt that they might just make it more difficult to understand the resolutions, which is what this page is about.

What do you think?

arcanis:
We should remove the globs from the documentation, they don't make sense.

- `foo`: will replace any instance of `foo`, anywhere
- `foo/bar`: will replace any instance of `bar` within dependencies called `foo`, anywhere

The glob patterns are unnecessary and mislead you into thinking that you can, for example, do something like "foo/bar/baz" - which isn't possible per Yarn's model (because `bar` is the same regardless of whether it's a dependency of `foo` or `qux`).

StephanBijzitter:
@arcanis 

This is not true, see the example above in the comments. Maybe it's different for v2, but v1 does not operate the way you describe.

arcanis:
I know the v1, and the behaviour you describe doesn't have the effect you think. If you use the exact same range of jQuery in two different places (like the `foo` I mentioned), they'll share the overrides of `bar/**/jquery` even if only one of them is a dependency of `bar`. This is because our lockfile format doesn't support (by design) storing the same combination of name+range multiple times.

mikolaj-leszczynski:
@arcanis 
> The glob patterns are unnecessary and mislead you into thinking that you can, for example, do something like "foo/bar/baz" - which isn't possible per Yarn's model (because bar is the same regardless of whether it's a dependency of foo or qux).

So you are saying that yarn does not respect a resolution like `"foo/bar/baz": "1.0.0"` ? Do I get you correct ?

 I thought yarn would do: "I will install baz@1.0.0 for every chilld of `bar` when `bar` is a direct child of `foo`".

PR #974, opened 2019-06-24
docs(installations): use correct tar filename

makepanic:
Right now the `tar` snippet will fail as the filename is wrong.


Haroenv:
thanks!

Daniel15:
Good catch, thanks for fixing it!

PR #975, opened 2019-07-02
chore(lint-staged): Upgrade lint-staged to upgrade js-yaml

BYK:
We have a security audit warning on `js-yaml` which is indirectly
included by `lint-staged`. Upgrading `lint-staged` to upgrade this
version. Although this is a semver-major change, the breaking
changes should not be affecting this project. If they do, I'll
resort back to selective resolutions to only upgrade `js-yaml`.


Haroenv:
thanks, this works!

PR #977, opened 2019-07-10
Adding yarn update info to docs

votemike:
`yarn install` helpfully says
```
warning Your current version of Yarn is out of date. The latest version is "1.16.0", while you're on "1.15.2".
info To upgrade, run the following command:
$ curl --compressed -o- -L https://yarnpkg.com/install.sh | bash
```
Yet when looking at the self-update documentation it doesn't actually tell you how to update yarn.


Haroenv:
good call!

Daniel15:
The instructions displayed actually differ based on the installation method you used to install Yarn. For example, if you install it using the Debian package, it will suggest using `apt-get`, and if you install it using Homebrew, it'll suggest the Homebrew commands to update the package.

votemike:
I'll update

votemike:
What should I update it with?
I was thinking a note sating that the user could look at the [install page](https://yarnpkg.com/en/docs/install) as this seems to detect the system and suggest upgrade options. And then all suggest running `yarn install` to get the correct command.
Or I could implement a `yarn self-update` command which says something like `This command is not implemented, please run ${getUpdateCommand()}`

PR #978, opened 2019-07-10
Updating docs surrounding the yarn audit level flag

votemike:
It's not clear that using `--level` on the `yarn audit` command does not affect the exit status of the command.
See https://github.com/yarnpkg/yarn/issues/7260


Haroenv:
good call!

PR #982, opened 2019-07-16
link to CHANGELOG from stable version link

JaKXz:
Since the release notes are no longer under the releases, I think it's much more helpful to see the release notes when clicking on the stable version. I'm not sure if it's worth it to add the fragment to the url with the current version since you'd have to change `1.17.3` to `1170` for example - not sure how consistent that's guaranteed to be.

### DIFF
--- a/_includes/version.html
+++ b/_includes/version.html
@@ -2,7 +2,7 @@
   <div>
     {{i18n.site_nav_stable_version}}:
     <strong class="navbar-text">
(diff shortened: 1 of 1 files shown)


Daniel15:
@arcanis - I wonder if the release scripts should automatically take the entries from the changelog and use those as the notes for the GitHub release.

JaKXz:
This is still an issue that could be solved, any updates @Daniel15? :)

JaKXz:
cc @arcanis

PR #984, opened 2019-07-24
Add json option to workspaces docs

ondrejsevcik:
There is missing `--json` option for `yarn workspaces info` command.


Haroenv:
thanks!

PR #988, opened 2019-08-03
feat(search): add refinements

Haroenv:
cc @orta, I still need to do the styling, but I know you were interested in exploring the different TS support over the index

### DIFF
--- a/js/src/components/Results.js
+++ b/js/src/components/Results.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import {
   Pagination,
   CurrentRefinements,
@@ -7,7 +7,7 @@ import {
   createConnector,
 } from 'react-instantsearch-dom';
 
-import { Hit } from './';
+import { Hit, Refinements } from './';
 import { isEmpty } from '../util';
 
 const body = document.querySelector('body');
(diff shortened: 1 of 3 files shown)


js/src/components/Results.js
```diff
@@ -38,8 +61,23 @@ const ResultsFound = ({ pagination, onTagClick, onOwnerClick }) => (
...
       />
+      <CurrentRefinements
+        transformItems={items => {
+          return [
+            ...new Map(
+              items.map(item => {
+                if (item.attribute === 'owner.name') {
+                  item.label = 'owner: ';
+                  return [item.attribute, item];
+                }
```
Haroenv:
```suggestion
                }
                if (item.attribute === 'types.ts' {
                  item.label = 'TypeScript: ';
                  return [item.attribute, item];
                }
```

Daniel15:
Interesting idea.

The "deprecated" facet is a bit odd though... I would have just expected it to be a boolean, rather than showing deprecation messages:

In this case I was just searching for "React".

Haroenv:
Yes, I was looking at how to show it properly, maybe need to change the index to properly index it, because you want three states really:

1. only  deprecated packages
2. no deprecated packages
3. all packages

I'll probably leave it out before merging, and add it later

orta:
One potential way to not need the extra modal UI is to have a smaller summary sub-heading with clickable links that do some of the more major refinements. For example the origin CocoaPods search did something like this:

Screen Shot 2019-08-04 at 9 37 48 AM

So that could be "Show only: *With Definitions* (123), *Not deprecated* (120), *By TJ Holowitz* (30)

Haroenv:
interesting idea! I think I'm mostly done playing around here, but we can indeed consider other UIs as well

sarahdayan:
Nice one! Here are a few thoughts:

- For discoverability, I'd show the refinements by default. Users can still hide it and you can save that as a preference, but it seems there's enough screen real estate to show it on load.
- The up/down icons in the **show/hide refinements** are misleading, I'd recommend going at least for left/right if you keep the arrow icons.
- The **TypeScript support** and **Deprecated** refinements could be displayed differently since they have a finite number of facets. Maybe displaying it inline, on top of the hits, with a different UI (toggle switches, icons?) would work well.

Haroenv:
talked to @sarahdayan, a solution for the icon is:

1. show the sidebar where there's place for it
2. don't show the button when it fits side-by-side

That way the arrows still make sense

PR #989, opened 2019-08-06
fix Chinese language code

justjavac:


Haroenv:
Hi @justjavac, thanks for  your contribution, this definitely looks correct, but I'm not fluent in Chinese languages, so I can't properly review. I'm going to ask around to see if I can verify the info

cht8687:
this fix is correct.

Haroenv:
thanks!

hax:
It seems `zh-sg` should still under `zh-Hans`.

cht8687:
thanks for correcting @hax 
didn't know zh-sg is under zh-Hans.

Haroenv:
that was fixed by @wongmjane in #990 :)

PR #991, opened 2019-08-06
Correct upgrade scope in usage summary

jvmccarthy:
Add missing `--scope` parameter in usage summary for `yarn upgrade` for a scope. (Omitting `--scope` results in a no-op.) Also remove the optional trailing forward slash (`/`) after the scope name for consistency as the example clarifies what a scope is.


Haroenv:
that seems correct, thanks!

Issue #968, opened 2019-06-16
Installation Guide leads to wrong yarn


cawoodm:
Following the official guide on Ubuntu 18.04 led to the "wrong" yarn being installed:
https://yarnpkg.com/lang/en/docs/install/#debian-stable

Executing `sudo apt remove cmdtest` before did not help. Apt-Get installs the wrong Yarn.

Following this guide worked:
https://linuxize.com/post/how-to-install-yarn-on-ubuntu-18-04/

The difference seems to be using "apt" instead of "apt-get":
```
sudo apt update
sudo apt install yarn
```

Daniel15:
Are you sure you ran `apt-get update` after adding the repo? `apt` and `apt-get` both use the same backend service so there shouldn't be a difference in how they behave. It's likely that the `apt update` is what helped, not just using `apt` vs `apt-get`.

Having said that, it's a good idea to update the site to use `apt` though.

PR #1013, opened 2019-10-14
change apt-get to apt in debian installation

clanghout:
In #968 @Daniel15 mentions `apt-get` should be updated to `apt` instead. I changed this in the debian install and nightly install instructions. 

The CI install docs for travis still have `apt-get` (see https://yarnpkg.com/lang/en/docs/install-ci/) I can also change those, but I believe `apt-get` is still preferred in scripts?


Daniel15:
Thanks!

PR #1023, opened 2019-12-02
install to add

Ron423c:
install to add next level var


Ron423c:
Minor fix to new standards of use

PR #1025, opened 2019-12-03
Adding yarn install --verbose to docs

LuisEnMarroquin:


Haroenv:
verbose is a global flag, but makes sense to specifically mention here

PR #1026, opened 2019-12-03
Updated the main page, as there is no information on the connection between the package manager and the language / technology for which it is used

proggga:
I spend much time, because I've searching usage of this package manager.
To be right, I want to know for which lang it uses. It's really good info, you can share code and so on, but I can't share my python/php/golang code through (Yea I can, but it will be very stupid)
This main/getting started has no any word about js. 

I found info about Yarn only from wiki about Npm 
image

### DIFF
--- a/lang/en/docs/getting-started.md
+++ b/lang/en/docs/getting-started.md
@@ -4,7 +4,7 @@ guide: docs_getting_started
 layout: guide
 ---
 
-Yarn is a package manager for your code. It allows you to use and share code
+Yarn is a package manager for your code. It allows you to use and share JavaScript code
 with other developers from around the world. Yarn does this quickly, securely,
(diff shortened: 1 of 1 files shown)


lang/en/docs/getting-started.md
```diff
@@ -4,7 +4,7 @@ guide: docs_getting_started
 layout: guide
 ---
 
-Yarn is a package manager for your code. It allows you to use and share code
+Yarn is a package manager for your code. It allows you to use and share JavaScript code
```
Haroenv:
```suggestion
Yarn is a package manager for your code. It allows you to use and share (e.g. JavaScript) code
```

You can share many more types of code, but I understand the main goal in JS

Issue #1027, opened 2019-12-03
Update install instructions for fish shell


mtjn:
I couldn't open a pull request because I couldn't find the file to edit for this page: https://yarnpkg.com/en/docs/install#debian-stable

The recommended technique for setting PATH in fish is to use `fish_user_paths` like so:

`set -U fish_user_paths (yarn global bin) $fish_user_paths`

c.f. https://fishshell.com/docs/current/tutorial.html#tut_path and https://wiki.archlinux.org/index.php/Fish#Setting_fish_as_default_shell

Thanks!

Haroenv:
the source is here https://github.com/yarnpkg/website/blob/master/lang/en/docs/_installations/unix_path_setup.md

PR #1028, opened 2019-12-03
Update fish shell path instructions

mtjn:
Resolves https://github.com/yarnpkg/website/issues/1027


Haroenv:
definitely clearer,tahnks!

PR #1030, opened 2019-12-06
install.sh: Post-verification cleanup (terminate gpg-agent)

### LINKED ISSUE #1029: install.sh script may not clean up gpg-agent processes
During the `install.sh` script, GPG is used to verify the integrity of the downloaded release, which is great.

There are some [reports](https://travis-ci.community/t/build-doesnt-finish-after-completing-tests/288) of stalled builds in Travis CI when `yarn` is installed during build process - particularly on the **windows** platform, but [potentially](https://github.com/GoogleChrome/lighthouse/pull/10075) also with **linux** platforms.

In many environments, spinning up a `gpg-agent` process is harmless and won't cause any problems after installation completes, but for Travis the fact that processes are still running is probably causing the build to 'wait' until all steps are complete (which in fact, they are - the `gpg-agent` process is just a side effect).

It looks like `gpgconf --kill gpg-agent` is [one possible way](https://superuser.com/questions/1075404/how-can-i-restart-gpg-agent#1150399) to clean up the daemon once it's no longer required.  Perhaps this can be added after the [verify](https://github.com/yarnpkg/website/blob/ad15aa4735bb4cfb79221bf583aeae22c7e30ce4/install.sh#L73) command runs, so that the process is removed once no longer required.

jayaddison:
This changeset terminates any `gpg-agent` processes created as side-effects during the `install.sh` script.

If _existing_ `gpg-agent` processes were found before GPG functionality is invoked in the script, those existing processes will be left untouched so as not to affect unrelated system/user functionality.

Fixes #1029

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -60,8 +60,17 @@ yarn_verify_integrity() {
   fi
 
   printf "$cyan> Verifying integrity...$reset\n"
+
+  # Check whether a gpg-agent process already exists for this session
+  existing_gpg_agent=$(pgrep gpg-agent || echo)
+
   # Grab the public key if it doesn't already exist
   gpg --list-keys $gpg_key >/dev/null 2>&1 || (curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --import)
+
+  # If no gpg-agent process was running previously, terminate any that may have been created during key import
+  if [ -z ${existing_gpg_agent} ]; then
+    gpgconf --kill gpg-agent
+  fi
 
   if [ ! -f "$1.asc" ]; then
     printf "$red> Could not download GPG signature for this Yarn release. This means the release can not be verified!$reset\n"
@@ -76,9 +85,6 @@ yarn_verify_integrity() {
     printf "$red> GPG signature for this Yarn release is invalid! This is BAD and may mean the release has been tampered with. It is strongly recommended that you report this to the Yarn developers.$reset\n"
     yarn_verify_or_quit "> Do you really want to continue?"
   fi
-
-  # Cleanup verification resources
-  yarn_verification_cleanup
 }
 
 yarn_link() {
@@ -208,17 +214,12 @@ yarn_install() {
   yarn_reset
 }
 
-yarn_verification_cleanup() {
-  gpgconf --kill gpg-agent
-}
-
 yarn_verify_or_quit() {
   read -p "$1 [y/N] " -n 1 -r
   echo
   if [[ ! $REPLY =~ ^[Yy]$ ]]
   then
     printf "$red> Aborting$reset\n"
-    yarn_verification_cleanup
     exit 1
   fi
 }
--- a/install.sh
+++ b/install.sh
@@ -62,14 +62,14 @@ yarn_verify_integrity() {
   printf "$cyan> Verifying integrity...$reset\n"
 
   # Check whether a gpg-agent process already exists for this session
-  existing_gpg_agent=$(pgrep gpg-agent || echo)
+  existing_gpg_agent=$(pgrep gpg-agent || pidof gpg-agent || echo)
 
   # Grab the public key if it doesn't already exist
   gpg --list-keys $gpg_key >/dev/null 2>&1 || (curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --import)
 
   # If no gpg-agent process was running previously, terminate any that may have been created during key import
   if [ -z ${existing_gpg_agent} ]; then
-    pkill gpg-agent || true
+    gpgconf --kill gpg-agent
   fi
 
   if [ ! -f "$1.asc" ]; then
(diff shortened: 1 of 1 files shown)


jayaddison:
cc @BanzaiMan - this may help with the issues re: GPG subprocesses you've seen user feedback about (ref: https://github.com/travis-ci/travis-ci/issues/8082#issuecomment-522590424).  It's important to make sure this PR doesn't cause side-effects elsewhere though.

jayaddison:
`gpg-connect-agent --no-autostart` also looked like a possibility to determine if the agent is current running.

However, trying that with version `2.2.12` of `gpg-connect-agent`, the agent is still started by a [reset](https://github.com/gpg/gnupg/blob/b966a7c142ab6341f4149d55e2609dbb9914acec/common/asshelp.c#L569-L571) process (confirmed via strace), so it may not be reliable.

install.sh
```diff
@@ -60,9 +60,18 @@ yarn_verify_integrity() {
   fi
 
   printf "$cyan> Verifying integrity...$reset\n"
+
+  # Check whether a gpg-agent process already exists for this session
+  existing_gpg_agent=$(pgrep gpg-agent || echo)
```
jayaddison: This is the best solution I could come up so far; essentially to use what is hopefully a relatively cross-platform command to determine whether a `gpg-agent` process already exists before key import is (potentially) attempted.  If previous agents did *not* exist, we can go ahead and remove any which were created.

`pgrep` has been available on Mac OS X since release 10.8 (released July 2012), and is AFAIK available by default in most Linux distributions.

Daniel15:
Seems reasonable to me. Does `pgrep` come preinstalled on minimal installs of popular Linux distros (Debian, Ubuntu, CentOS)?

jayaddison:
@Daniel15 Good question, perhaps there's an authority / comparison on which Linux distros it's available under; I'm taking a bit of a look around just now.

jayaddison:
While searching around elsewhere, here's whether `pgrep` is available in each of the latest Docker Hub official images for Centos, Debian and Ubuntu:

CentOS
```
sh-4.4# cat /etc/centos-release
CentOS Linux release 8.0.1905 (Core) 
sh-4.4# whereis pgrep
pgrep: /usr/bin/pgrep
```

Debian:
```
# cat /etc/debian_version
10.2
# which pgrep
#
```

Ubuntu
```
# cat /etc/debian_version
buster/sid
# which pgrep
/usr/bin/pgrep
#
```

So Debian does **not** seem to have `pgrep` by default.

jayaddison:
@Daniel15 Hrm.  I'm not finding many resources to compare command-line tooling per-distro; it'd probably take a bit more research.  Something like [caniuse](https://caniuse.com/) for command-line tools would be wonderful.. :pray: 

One option to handle cases like Debian could be to fall-back to alternatives like the `killall` command.  It doesn't look like `yarnpkg` currently supports Solaris or other non-Linuxes where that command can be a bit more [dangerous](https://linux.die.net/man/1/killall), so perhaps that's fine (or even a better alternative than using `pkill` in the first place).

Whatever approach, I'm keen to be a bit cautious, since this script seems fairly important infrastructure-wise :)

jayaddison:
@Daniel15 @BanzaiMan Just for transparency; it looks [likely that this wasn't the true root cause](https://github.com/GoogleChrome/lighthouse/pull/10075#issuecomment-562808911) of Travis CI timeout issues seen in `lighthouse` during Linux builds.

It still seems useful to clean-up non-install side-effects of the script, so there may still be some value here, but I think the importance & impact of this PR are low.

Glad for any thoughts, and I'll gradually continue on to see if there are more cross-platform alternatives to `pgrep` and `pkill`, but there's no urgency (and I don't think it's a magic solution to any other customer issue reports).

jayaddison:
Investigating `pidof` and `killall` as alternatives to `pgrep` and `pkill`, respectively:

Alpine
```
/ # cat /etc/alpine-release 
3.10.3
/ # command -v pidof
/bin/pidof
/ # command -v killall
/usr/bin/killall
/ #
```

Centos
```
sh-4.4# cat /etc/centos-release
CentOS Linux release 8.0.1905 (Core) 
sh-4.4# command -v pidof
/usr/sbin/pidof
sh-4.4# command -v killall
sh-4.4#
```

Debian
```
# cat /etc/debian_version
10.2
# command -v pidof
/bin/pidof
# command -v killall
# 
```

Ubuntu
```
# cat /etc/debian_version
buster/sid
# command -v pidof
/bin/pidof
# command -v killall
# 
```

Findings:

- `pidof` is a widely available alternative as a fallback from `pgrep`.  Note that `pidof` is **not** available on Mac OS X, so we should still try both.
- `killall` is not available widely; it may be better to rely on `gpgconf --kill gpg-agent` for `gpg-agent` termination since it is bundled with the `gpg` toolset

jayaddison:
@Daniel15 This turned out not to be the cause of the Travis build failures I was seeing; instead it looks like that was caused by slow cache restoration - so I'll go ahead and close this PR.

Daniel15:
Thanks for the update... Feel free to ping me again if you'd like to reopen this at some point.

PR #1033, opened 2019-12-29
Update global.md

mtourj:
Create-react-app no longer supports global installs, so this may be misleading. A note like the one I added or a change of the example would be beneficial to the usefulness of this page.

### DIFF
--- a/lang/en/docs/cli/global.md
+++ b/lang/en/docs/cli/global.md
@@ -14,17 +14,15 @@ layout: guide
 
 _Note: Unlike the `--global` flag in npm, `global` is a command which must immediately follow `yarn`. Entering `yarn add global package-name` will add the packages named `global` and `package-name` locally instead of adding `package-name` globally._
 
-This is useful for developer tooling that is not part of any individual project but instead is used for local commands. One such example is [create-react-app](https://github.com/facebookincubator/create-react-app) which can be installed globally like this:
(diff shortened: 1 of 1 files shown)


Daniel15:
I don't think this is accurate. Create React App must be globally installed. Their instructions use npx which also uses global installs.

mtourj:
When you try to run 'create-react-app <name>' on a globally installed, outdated version of create-react app, you get the following message:

`A template was not provided. This is likely because you're using an outdated version of create-react-app.
Please note that global installs of create-react-app are no longer supported.`

Anytime you run `npx create-react-app` the latest version is automatically installed, ran, and deleted after the process is complete.

mtourj:
> I don't think this is accurate. Create React App must be globally installed. Their instructions use npx which also uses global installs.

Found in the React docs:
```If you've previously installed create-react-app globally via npm install -g create-react-app, we recommend you uninstall the package using npm uninstall -g create-react-app to ensure that npx always uses the latest version.```

https://create-react-app.dev/docs/getting-started/

mtourj:
Also running a globally installed version of `create-react-app` will eventually yield the following message when a new version becomes available:
`A template was not provided. This is likely because you're using an outdated version of create-react-app.
Please note that global installs of create-react-app are no longer supported.`

Haroenv:
I think we should use a different example, that _is_ well suited for global installs, but I can't think of any. Do you know one?

mtourj:
You can use nodemon as an example

Haroenv:
that's a good example, thanks!

PR #1035, opened 2020-01-08
Update debian nodesource link

gricey432:
The anchor changed on the nodejs site


Daniel15:
This should probably link to https://github.com/nodesource/distributions/blob/master/README.md#deb instead, as the Node.js site simply links to that repo.

gricey432:
I did think that as I was following the links myself, but not sure if that's as stable / didn't want to change too much as a new contributor. Happy to change the PR if that's what you'd prefer

PR #1040, opened 2020-01-26
Remove duplicate meta viewport property

cone56:
`initial-scale=1` unnecessary appears twice in the meta viewport tag. This PR deletes the duplicate.


Haroenv:
thanks!

PR #1041, opened 2020-02-04
Add mention to ignore .pnp.* files

foxyblocks:
As as a user that is new to PnP but not to yarn who is looking to migrate an existing project, I felt confused about this point when using the getting started docs. I think it would be helpful to add here.


Haroenv:
might be confusing to link the v2 docs here, but we don't have v1 docs on this.

PR #1050, opened 2020-03-15
Update opensearch.xml

c0b41:
for the new yarn url.


Daniel15:
Should we move this to the new site instead?

Haroenv:
Yes, would need to check if the new site has this. I think this PR is fine, but we probably  need to add the equivalent in berry

c0b41:
can we merge this? i can't use opensearch this reason

PR #1081, opened 2020-10-01
Update

Gauravrathi1122:


Haroenv:
Thanks for your pr, but the original text reads better to me. Good luck with more substantial pull requests in the future though!

PR #1088, opened 2020-10-01
tableRename -debug.html to -debug.htmlcss

Ayushi1103:


unnown:
Please mark as invalid and spam, this is an obvious attempt to get a free t-shirt by 'contributing'(hacktoberfest)
This will also ban the user for trying to abuse the system by doing marginal effort

Haroenv:

PR #1090, opened 2020-10-01
Explicitly use .yarn directory as cache directory

RealOrangeOne:
This saves caching the whole of `node_modules`, but still provides a performance improvement to builds, as an offline cache can be used.

The default cache location isn't persisted by GitLab

### DIFF
--- a/lang/en/docs/_ci/gitlab.md
+++ b/lang/en/docs/_ci/gitlab.md
@@ -21,13 +21,12 @@ before_script:
   - export PATH="$HOME/.yarn/bin:$HOME/.config/yarn/global/node_modules/.bin:$PATH"
 ```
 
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
Looks like the deploy fails are unrelated to this PR:

> CROWDIN_API_KEY is undefined.

Haroenv:
I'm not sure why you're suggesting not to cache node_modes, not having to fetch for subsequent builds still seems useful to me

RealOrangeOne:
In my experience `node_modules` is almost always larger than the download cache, and so there's more to cache, more to change, which slows down builds.

This also matches how GitLab recommend caching for NPM directly: https://docs.gitlab.com/ee/ci/caching/#caching-nodejs-dependencies

PR #1108, opened 2021-01-05
docs: add update notices

paul-soporan:
The time for us to add some v2+ upgrade notices has finally come.

This PR adds:
- a migration banner
- a migrate button on the homepage
- updates the installation page to recommend installation via npm
- updates the self-update docs a bit
- adds an update notice to the installation page
- a few other small changes that I've discussed with @arcanis and the rest of the team

### DIFF
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -16,7 +16,7 @@ site_nav_blog: Blog
 site_nav_stable_version: Classic Stable
 site_nav_rc_version: Classic Release Candidate
 site_nav_nightly_version: Classic Nightly
-site_nav_node_support: Classic Node
+site_nav_node_support: Node
 
 site_bsd_license: Distributed under BSD License
 site_code_of_conduct: Code of Conduct
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -13,10 +13,10 @@ site_nav_getting_started: Getting Started
 site_nav_documentation: Docs
 site_nav_packages: Packages
 site_nav_blog: Blog
-site_nav_stable_version: Stable
-site_nav_rc_version: Release Candidate
-site_nav_nightly_version: Nightly
-site_nav_node_support: Node
+site_nav_stable_version: Classic Stable
+site_nav_rc_version: Classic Release Candidate
+site_nav_nightly_version: Classic Nightly
+site_nav_node_support: Classic Node
 
 site_bsd_license: Distributed under BSD License
 site_code_of_conduct: Code of Conduct
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -57,6 +57,7 @@ homepage_feature_reliable_image_alt: Watercolour of cat waving from seat behind 
 homepage_c2a_text: What are you waiting for?
 homepage_c2a_button: Get Started
 homepage_install_button: Install Yarn
+homepage_migrate_button: Migrate to Yarn 2+
 
 homepage_featurette_offline_title: Offline Mode
 homepage_featurette_offline_description: |
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -5,6 +5,10 @@
 site_title: Yarn
 site_description: Fast, reliable, and secure dependency management.
 
+site_news_important: Important
+site_news_docs_cover_v1: This documentation covers Yarn 1 (Classic).
+site_news_see_v2_docs: For Yarn 2+ docs and migration guide, see yarnpkg.com.
+
 site_nav_getting_started: Getting Started
 site_nav_documentation: Docs
 site_nav_packages: Packages
--- a/_data/i18n/en.yml
+++ b/_data/i18n/en.yml
@@ -267,7 +267,9 @@ users_add_requirements_logo: >
 
 install_intro: >
   Before you start using Yarn, you'll first need to install it on your system.
-  There are a growing number of different ways to install Yarn:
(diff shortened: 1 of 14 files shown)


Daniel15:
It looks like the build failed for some reason:

```
12:16:58 PM: Error running command: Build script returned non-zero exit code: 2
```
https://app.netlify.com/sites/yarnpkg/deploys/5ff4c8e1102fa80007bb0725

Can you please take a look? I'd like to look at the preview site while reviewing.

paul-soporan:
🤔 I also noticed that but then thought nothing of it as the CI has been consistently failing since June in most pending and merged PRs with exactly the same (mysterious) reason:

```
10:13:40 PM: Error running command: Build script returned non-zero exit code: 2
10:13:40 PM: Failing build: Failed to build site
10:13:40 PM: failed during stage 'building site': Build script returned non-zero exit code: 2
```

It's weird that the logs don't include any actual error.

image

Haroenv:
Netlify has changed the behaviour in June where api keys / env variables are no longer given to non-team deploys, I think that's what causes the issue

Haroenv:
Hi, I think I've fixed the deploy script, could you rebase from master?

paul-soporan:
Rebased and the website deployed successfully. 👍

_data/i18n/en.yml
```diff
@@ -5,14 +5,18 @@
...
 site_nav_packages: Packages
 site_nav_blog: Blog
-site_nav_stable_version: Stable
-site_nav_rc_version: Release Candidate
-site_nav_nightly_version: Nightly
-site_nav_node_support: Node
+site_nav_stable_version: Classic Stable
+site_nav_rc_version: Classic Release Candidate
+site_nav_nightly_version: Classic Nightly
+site_nav_node_support: Classic Node
```
Haroenv, line 19:
```suggestion
site_nav_node_support: Node
```

Haroenv (approved):
I think overall this change is not too "in-your-face" and makes sense

_includes/version-with-upgrade-notice.html
```diff
@@ -0,0 +1,10 @@
+<div class="alert alert-warning news-container-alert mt-4">
+  <a class="news-overlay" href="https://yarnpkg.com/getting-started/migration#why-should-you-migrate"></a>
+  <div class="news-inner">
+    These instructions only cover Yarn versions prior to 2.0. They entered <a href="https://en.wikipedia.org/wiki/Maintenance_mode">maintenance mode</a> in January 2020 and will eventually reach their end-of-life in terms of support. To see the up-to-date documentation, check the main website: <a href="https://yarnpkg.com/getting-started">yarnpkg.com/getting-started</a>.
+    <br>
+    The latest Yarn version is: <img alt="Latest CLI Release" src="https://img.shields.io/npm/v/@yarnpkg/cli/latest?label=latest">
```
arcanis, line 6: I think we should point to the migration guide rather than getting-started, since people accessing this install page will likely already know what Yarn is, and will be more interested by either the install instructions or migration guide.

```suggestion
    <div>
      These instructions only cover Yarn versions prior to 2.0. Those versions entered <a href="https://en.wikipedia.org/wiki/Maintenance_mode">maintenance mode</a> in January 2020 and will eventually reach their end-of-life in terms of support. Please see the main website for the most up-to-date documentation: <a href="https://yarnpkg.com/getting-started/migration#why-should-you-migrate">yarnpkg.com/getting-started/migration</a>.
    </div>
    <div style="margin-top: 15px">
      The latest Yarn version is: <img alt="Latest CLI Release" src="https://img.shields.io/npm/v/@yarnpkg/cli/latest?label=latest">
    </div>
```

arcanis (approved):

_includes/version-with-upgrade-notice.html
paul-soporan: I was initially thinking about putting the migration guide, but then in the notice you wrote for me to put on the website, you put the getting-started page and after some thought, I managed to convince myself that it's fine. :P 

Since the entire notice already has an overlay to `https://yarnpkg.com/getting-started/migration#why-should-you-migrate`, I think we should rather link the v2+ installation page instead.

What do you think?

arcanis: It's likely most people won't see that the overlay is clickable - having an explicit link is still a good idea, imo.

As for the target, people on the Classic website will likely be in a state where they want to work on a Classic project - by which point knowing how to install Yarn 2 will matter less than knowing how to migrate it altogether 🤔

Daniel15:
Thanks for fixing the preview @Haroenv!

@paul-soporan  I think this PR makes sense to me. 

What's the reason for moving to using npm to install Yarn rather than continuing to promote the native installation methods (Windows installer, Debian package, etc)?

Also, does this mean that there'll be no more releases of Yarn v1? I have a build server that's only used for updating the Chocolatey and Homebrew packages - it sounds like I can shut that down now?

arcanis:
> What's the reason for moving to using npm to install Yarn rather than continuing to promote the native installation methods (Windows installer, Debian package, etc)?

People often have problems where they try to install Yarn one way and upgrade it another. Generally speaking supporting all those stores adds moving pieces, and I'd like to simplify the infra to keep things as simple as possible in the main release pipeline.

There's also that npm is installed on all Node machines, so the install instructions are easier to follow: no need to select the right system, etc. Since it works just fine, apart from the slightly awkward "install Chrome with IE" situation, it seems reasonable to focus on that. 

For the build server it's as you wish - the test / build pipeline actually broke some time ago so I had to make manual publishing anyway.

PR #1133, opened 2021-09-11
yarnpkg.com --> classic.yarnpkg.com

mrienstra:
See yarnpkg/berry#3425 for motivation. This will update `<link rel="canonical">`, so Google Search will show "classic" URL. Also updates `<meta property="og:url">`.

`_config.yml `: This change will add `classic` subdomain to `<link rel="canonical">` & `<meta property="og:url">` for most pages, via [_layouts/default.html](https://github.com/yarnpkg/website/blob/5be62ee/_layouts/default.html)

`js/src/components/Details.js`: This change will update `<link rel="canonical">` & `<meta property="og:url">` for  "Package detail" pages, e.g. https://classic.yarnpkg.com/en/package/Babel

Note: I haven't yet attempted to test these changes locally. I'm willing to do so if there is any uncertainty as to the outcome. Just took a look at the "deploy preview" and it looks good.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -3,7 +3,7 @@ title: Yarn
 description: > # this means to ignore newlines until "baseurl:"
   Fast, reliable, and secure dependency management.
 baseurl: ""
-url: "https://yarnpkg.com"
+url: "https://classic.yarnpkg.com"
 twitter_username: yarnpkg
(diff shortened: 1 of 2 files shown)


Daniel15:
Looks good to me. Thanks!

PR #1137, opened 2021-10-03
update discord logo

treboryx:


Haroenv:
Thanks, the logo has indeed updated!

treboryx:
> Thanks, the logo has indeed updated! Have you made a similar PR to yarnpkg/berry?

No, I'll check if update is needed there as well.

PR #1176, opened 2022-11-14
Support linking with older yarn versions

achempion:
We have two project which use yarn, the website uses yarn 1 and the package uses yarn 3.

We want to be able to test & develop package locally and have live updates for the main react app.

I found this pease missing how to support integration with legacy projects.

[related issue](https://github.com/yarnpkg/berry/issues/90)


philippmalkov:
Hi! Excuse me, which version of Yarn v3 do you use? I'm using v3.2.4 and it says:

```
> YARN_IGNORE_PATH=1 yarn link
Unknown Syntax Error: Not enough positional arguments.
```

achempion:
@philippmalkov we're on `3.2.0`

PR #1177, opened 2022-11-14
Add better entry point description

jircdev:
This change fixes the missing flag in the first example. If you just run yarn version it will only plot the version, not change it.
Update the note from the end of the first example to the --new-version flag where it is related to.

The current docs say the next about NodeJS:
>   is used to define the entry point of your code used by programs like Node.js. If unspecified it will default to index.js.

This PR proposes to change the NodeJS definition, seeking to avoid confusion about how javascript and javascript packages really work. 

Also,  today NodeJS purpose the "exports" entry point usage instead of "main" entry, and this entry point (exports) has better support for universal packages.

Also is proposed to add a mention of "exports" entry point which is taken over "main" entry point if it is defined. The paragraph is added after the explanation about the entry points contained.


Haroenv:
Good little addition, thanks!

PR #1183, opened 2023-01-04
Remove Font Awesome

nickmccurdy:
I noticed Font Awesome was out of date, but in trying to update it I couldn't actually find any usages of its CSS, so I simply removed it.


Haroenv (approved):
I can't find any usages either, must have been removed in a different refactor. Thanks!

PR #1208, opened 2023-08-31
Update run.md

multistarmedia:
changed typo from 'jest' to 'test'


Haroenv:
No, the command is called `jest`, as that's what's in `package.json#scripts#test`. thanks for your contribution!

PR #1209, opened 2023-09-06
Update twitter logo

priyanshuchaudhary53:
This PR updates the old Twitter logo in the header to the new one.

## Before
before

## After
after


Haroenv:
There's no need for that IMO

PR #1213, opened 2023-09-21
change --global to --location=global

amir-reza-tavakkoli:
Hi,
I noticed that npm has recently deprecated the --global flag, so I changed the relevant occurrences of that.


Haroenv:
Thanks, but if we make this change, it will only work for newer version of npm, while --global is still supported. Maybe we should wait a bit?

amir-reza-tavakkoli:
> Thanks, but if we make this change, it will only work for newer version of npm, while --global is still supported. Maybe we should wait a bit?

I guess you are right, didn't notice that.

PR #1227, opened 2024-06-17
ci: add Azure Static Web Apps workflow file

milyes:
on-behalf-of: @Azure opensource@microsoft.com


Haroenv:
as far as I can tell nobody has asked for this, so I won't accept a pull request that reads credentials

PR #1245, opened 2025-10-21
apt-key no longer exists in ubuntu 25

RubenGarcia:


Daniel15:
I merged the other PR. Do we still need this one? The new way of doing it should work for outdated Debian and Ubuntu versions too.

RubenGarcia:
Ok, if the new way of doing things works in older distros too, this is not needed. Closing.

PR #1250, opened 2026-03-16
Update yarn link instructions for Yarn Modern projects

clemyan:
Updates the instructions for linking a Yarn Modern project into Yarn Classic to account for Corepack usage and Yarn Classic's own Corepack check

See https://github.com/yarnpkg/berry/issues/5339#issuecomment-4040549027


Daniel15:
Thanks!

